PR #136, opened 2014-12-30
fix error: variable 'mem' set but not used [-Werror=unused-but-set-varia...

alagoutte:
...ble]


tatsuhiro-t:
Thank you! Merged now.

Issue #87, opened 2014-10-07
O_CLOEXEC not defined


acesso:
After few hours compiling all the tool chain to compile it on Centos 5.5 i´ve found out that util.cc uses O_CLOEXEC witch is not on this distro kernel since this function was added on  2.6.23 (centos5 still uses 2.6.18). I know this is old but may address builders on old boxes.

barracuda156:
Missing `O_CLOEXEC ` supposed to be fixed here: https://www.gnu.org/software/gnulib/manual/html_node/fcntl_002eh.html

PR #142, opened 2015-01-22
Use fcntl and FD_CLOEXEC if O_CLOEXEC is undefined

acesso:
Same reported at #87 but at src/shrpx_config.cc src/instead of util.cc

### DIFF
--- a/src/shrpx_config.cc
+++ b/src/shrpx_config.cc
@@ -251,7 +251,7 @@ FILE *open_file_for_write(const char *filename) {
  
     // We get race condition if execve is called at the same time.
     if(fd != -1) {
-        fcntl(fd, F_SETFD, FD_CLOEXEC);
+        _make_socket_closeonexec_(fd);
     }
   #endif
   if (fd == -1) {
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Thank you for PR!  Could you check my comment?

acesso:
fixed and tested, compiled like a charm

src/shrpx_config.cc
```diff
@@ -242,8 +242,18 @@ read_tls_ticket_key_file(const std::vector<std::string> &files) {
...
+  #if defined O_CLOEXEC
+    auto fd = open(filename, O_WRONLY | O_CLOEXEC | O_CREAT | O_TRUNC,
+                     S_IRUSR | S_IWUSR);
+ #else
+    auto fd = open(filename, O_WRONLY | O_CREAT | O_TRUNC,
+                   S_IRUSR | S_IWUSR);
+ 
+    // We get race condition if execve is called at the same time.
+    if(fd != -1) {
+        int make_socket_closeonexec_(fd);
```
tatsuhiro-t: This is not correct to me.
I proposed specifically `util::make_socket_closeonexec(fd);`

tatsuhiro-t:
Merged via 7ea8037ee1a274cb450b0453736cc793c3022407
Thank you!

PR #145, opened 2015-01-28
nghttpx: set the supplementary group access list

B4dM4n:
When dropping privileges use [`initgroups(3)`](http://linux.die.net/man/3/initgroups) to set the supplementary group list of the selected user.
This allows to use the `ssl-cert` group on Ubuntu for example


tatsuhiro-t:
Thank you!  Merged via commit 243a813

PR #165, opened 2015-02-23
Adding address option to select the address to bind to

bcard:
This change adds a `-a` parameter that can be used to select the IP address that the http server will bind to.  During testing I used this option to use multiple http servers on the same machine each with different addresses.  The parameter is optional and if not specified then the server binds to the default address returned from `getaddrinfo`.  This option is available in other http servers such as node's `http-server`.  I also added some additional logic to the print statement so that the address is printed along with the port.

Note I was having some trouble with getting clang-format to work correctly on Ubuntu so I had to edit the formatting manually.  I think it is pretty close to the style that you have setup but let me know if there are any formatting changes that need to be addressed.

### DIFF
--- a/src/HttpServer.cc
+++ b/src/HttpServer.cc
@@ -31,6 +31,7 @@
 #include <fcntl.h>
 #include <netinet/in.h>
 #include <netinet/tcp.h>
+#include <arpa/inet.h>
 
 #include <cassert>
 #include <set>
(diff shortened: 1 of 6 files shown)


tatsuhiro-t:
Thank you for pull request.  I commented one thing.
For clang-format, I'll on your behalf squash the commits into one and perform clang-format and commit it (author is still you, of course) if you agree.

bcard:
That's fine with me thank you, I can look at addressing your comment over the next day.

tatsuhiro-t:
Probably it would be good to move numeric_name() in nghttp.cc to util.cc and share it with all applications.  Signature should be changed as `numeric_name(const struct sockaddr *sa, socklen_t salen)` however to avoid too getaddrinfo-ish.  But we don't too much ask for this in this PR.

bcard:
I refactored some of the code to move the `numeric_name` function into utils and call this from `nghttp.cc` and `HttpServer.cc`.  I missed your later comment and didn't change the function signature and it takes the whole `addrinfo` struct at the moment.  I can update that to finish this change up.

tatsuhiro-t:
Looks good.   I appreciate much if you update numeric_name signature like what I said.   If that is done, I'll merge this PR.  Thanks you.

bcard:
Ok I've made that last signature change.

tatsuhiro-t:
Thanks!  I'll merge this PR soon.

tatsuhiro-t:
Formatted using clang-format and merged.  I forgot to squash the commits, but it's OK.
When solving merge conflict, I removed unused function get_ip_str since we agreed to use util::numeric_name().
Thank you very much!

PR #186, opened 2015-03-16
fix for segfault by reserving correct worker count

icing:
Thanks for the hint, this works on my system correctly so far.


tatsuhiro-t:

PR #189, opened 2015-03-21
README.rst: OpenSSL 1.0.2 has already been released

Lekensteyn:
OpenSSL 1.0.2 is already released. Avoid the confusing wording that
seems to suggest that a development version of OpenSSL 1.0.2 provides
ALPN support.


tatsuhiro-t:
Thank you. Merged now.

Issue #195, opened 2015-03-27
Plain Windows MSVC build supported?


remoe:
Hi

The current build stack doesn't work with MSVC without install a Cygwin/MinGW stack. Why not support CMake at least for the nghttp2-lib?

tatsuhiro-t:
We have Makefile.msvc
https://github.com/tatsuhiro-t/nghttp2/blob/master/lib/Makefile.msvc

It seems outdated now, but if added missing source files, I think it should work.

remoe:
Thanks. I know, but it needs a lot of unix tools to build correct. (for example nm, grep, subst ...). CMake solves some issues without such shell tools. 

I've tried it with cygwin. I had errors at the end of ./configure, I've tried with VS2013:

```shell
...
checking consistency of all components of python development environment... no
configure: WARNING:
  Could not link test program to Python. Maybe the main Python library has been
  installed in some non-standard library path. If so, pass it to configure,
  via the LDFLAGS environment variable.
  Example: ./configure LDFLAGS="-L/usr/non-standard-path/python/lib"
  ============================================================================
   ERROR!
   You probably have to install the development version of the Python package
   for your distribution.  The exact name of this package varies among them.
  ============================================================================

checking for cython.py... no
checking for cython... no
checking whether cl.exe supports C++11 features by default... yes
checking whether std::chrono::steady_clock is available... yes
checking whether std::future is available... yes
checking whether std::map::emplace is available... yes
checking for library containing dlsym... no
checking for library containing clock_gettime... no
checking for ZLIB... yes
checking for CUNIT... no
configure: WARNING: No package 'cunit' found
checking for CU_initialize_registry in -lcunit... no
checking for ev_time in -lev... no
checking for OPENSSL... yes
checking for LIBEVENT_OPENSSL... no
configure: No package 'libevent_openssl' found
checking for JANSSON... no
configure: No package 'jansson' found
checking for xml2-config... /usr/bin/xml2-config
checking for libxml - version >= 2.7.7... no
*** Could not run libxml test program, checking why...
*** The test program failed to compile or link. See the file config.log for the
*** exact error that occured. This usually means LIBXML was incorrectly installed
*** or that you have moved LIBXML since it was installed. In the latter case, you
*** may want to edit the xml2-config script: /usr/bin/xml2-config
checking for library containing malloc_stats_print... no
checking for LIBSPDYLAY... no
configure: No package 'libspdylay' found
[...]

PR #196, opened 2015-03-27
Windows build fixes for #195

remoe:
- source file list updated
- zlib removed
- build of .def files removed

### DIFF
--- a/lib/Makefile.msvc
+++ b/lib/Makefile.msvc
@@ -2,15 +2,14 @@
 # GNU Makefile for nghttp2 / MSVC.
 #
 # By G. Vanem <gvanem@yahoo.no> 2013
+# Updated 3/2015 by Remo Eichenberger @remoe
 # The MIT License apply.
 #
 
 #
 # Choose your weapons:
-#  Set 'ZLIB_ROOT' to the root of zlib.
 #  Set 'USE_CYTHON=1' to build and install the 'nghttp2.pyd' Python extension.
 #
-ZLIB_ROOT  = g:/MingW32/src/Compression/zlib-1.2.8
 USE_CYTHON = 1
 
 _VERSION   := $(shell grep AC_INIT ../configure.ac | cut -d'[' -f3 | sed -e 's/-DEV], //g')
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Thank you! Merged now.

Elavarasan83:
Hello,
I'm compiling sources of Apache web server in windows x64 machine. Apache web server has dependency to nghttp2. I want to build nghttp2 in windows x64 machine. 

I have some questions:

1. Is the "Makefile.msvc' in 'lib' folder of nghttp2 is up to date? If not then what are are all the sections i have to update whenever I get a new version of nghttp2. I see 'nghttp2_rcbuf.c' is missing in 'NGHTTP2_SRC' section.

2. Apache expects .lib and not the .dll. I assume the output generated will have .lib file of nghttp2.

3.  In https://github.com/nghttp2/nghttp2, it is mentioned 'libevent' is necessary & also the macro '__STRICT_ANSI__' to be undefined. I think these steps are necessary only when we use cygwin. Am I right?

4.  The below mentioned nmake command is sufficient to generate the output or any othe configuration is required.
  **C:\nghttp2\lib>** _nmake /f Makefile.msvc_

Please clarify.

tatsuhiro-t:
> 1. Is the "Makefile.msvc' in 'lib' folder of nghttp2 is up to date? If not then what are are all the sections i have to update whenever I get a new version of nghttp2. I see 'nghttp2_rcbuf.c' is missing in 'NGHTTP2_SRC' section.

I suppose so.  I see nghttp2_rcbuf.c in Makefile.msvc in master branch

> 2 Apache expects .lib and not the .dll. I assume the output generated will have .lib file of nghttp2.

This Makefile was contributed by users, and I haven't executed it since I don't have the Windows development environment.  Did you see that it only produced dll, and not .lib file?

> 3 In https://github.com/nghttp2/nghttp2, it is mentioned 'libevent' is necessary & also the macro 'STRICT_ANSI' to be undefined. I think these steps are necessary only when we use cygwin. Am I right?

libevent is required to compile the programs under examples.  Makefile.msvc only compiles libnghttp2, so you don't need libevent.

I think STRICT_ANSI stuff is for cygwin, but I may be wrong.

> 4 The below mentioned nmake command is sufficient to generate the output or any othe configuration is required. The command will be run using to VS2013 x64 native tools command prompt.
C:\nghttp2\lib> nmake /f Makefile.msvc

The comment from Makefile.msvc says it is for "GNU Makefile for nghttp2 / MSVC", so it may require GNU Make.

Elavarasan83:
Dear tatsuhiro-t,
Thanks for the quick response. I will install GNU make and do the compilation

PR #198, opened 2015-03-31
Improve Dockerfile

wzyboy:
I love your idea of cross-compiling Android nghttpx in Docker! I do not have to make my ubuntu host a mess now.

Docker will generate a new layer for every KEYWORD (RUN, WORKDIR, etc), so more keywords means more layers and slower DeviceCreate() processes. So I made some improvements to the Dockerfile, which will make it builds faster and more efficiently.

Hope you like it :-)


tatsuhiro-t:
Looks good.  Merged.  Thank you!

PR #202, opened 2015-04-03
"static inline" fix for build with VS2013

remoe:
"static inline" doesn't build on VS2013:

https://msdn.microsoft.com/en-us/library/z8y1yy88.aspx


tatsuhiro-t:
Thank you!  Merged now.

PR #206, opened 2015-04-09
fix comma at end of enumerator list [-Wpedantic]

alagoutte:
Need also push on v1.0.0 ?


tatsuhiro-t:
Thank you! Merged now.

PR #211, opened 2015-04-21
Graceful shutdown and joinable server

sunxiaoguang:
Make it possible to shutdown server gracefully and also joinable.

### DIFF
--- a/src/asio_io_service_pool.cc
+++ b/src/asio_io_service_pool.cc
@@ -35,8 +35,6 @@
 //
 #include "asio_io_service_pool.h"
 
-#include <future>
-
 namespace nghttp2 {
 
 namespace asio_http2 {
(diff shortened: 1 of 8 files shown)


tatsuhiro-t:
Thank you for PR!
Clearly this breaks current API usage.   I suggest that we should preserve behavior of http2::listen_and_server (that is, block in listen_and_serve forever) and we can add detached version "listen_and_serve" where programmer can call http2::join() wherever they like.

It would be good to rebase the commits, since b2ce8ac is somewhat duplicated.

sunxiaoguang:
So I'm going to add a extra argument to specify if the server should be started in async mode or not. Do you prefer changing function signature to add an argument with default value or add a new method with one more argument. Adding new method doesn't break ABI, but add redundancy. Not sure which way you prefer. I would choose whatever the way you feel more comfortable with.

tatsuhiro-t:
I'm OK with adding parameter with default value.
asio lib is still not entirely stable right now, so ABI breakage is acceptable for now.

sunxiaoguang:
Sure, I will do it that way.

sunxiaoguang:
Committed, please review

tatsuhiro-t:
Thank you!  Reviewing now...

tatsuhiro-t:
Merged now.  Thank you very much.

sunxiaoguang:
It's my pleasure.

PR #217, opened 2015-04-29
fix comma at end of enumerator list [-Wpedantic]

alagoutte:


tatsuhiro-t:
Thank you! Merged now.

tatsuhiro-t:
Probably, we should add -Wpedantic to --ennable-werror mode

alagoutte:
Yes, you can try but always some "strange" error.. 
and for same case, it will be complicated to fix 
like 
nghttp2_session.c: In function 'session_close_stream_on_goaway':
nghttp2_session.c:2104:3: error: initializer element is not computable at load time [-Werror]
   nghttp2_close_stream_on_goaway_arg arg = {session, NULL, last_stream_id,
   ^
nghttp2_session.c:2104:3: error: initializer element is not computable at load time [-Werror]
nghttp2_session.c:2105:45: error: initializer element is not computable at load time [-Werror]
                                             incoming};

or 
client.c:448:29: error: initializer element is not computable at load time [-Werror]
                             MAKE_NV_CS(":path", req->path),

or...

../third-party/http-parser/http_parser.h:125:48: error: comma at end of enumerator list [-Werror=pedantic]
 #define XX(num, name, string) HTTP_##name = num,
                                                ^
../third-party/http-parser/http_parser.h:121:3: note: in expansion of macro ‘XX’
   XX(26, MKCALENDAR,  MKCALENDAR)   \
   ^
../third-party/http-parser/http_parser.h:126:3: note: in expansion of macro ‘HTTP_METHOD_MAP’
   HTTP_METHOD_MAP(XX)
   ^

tatsuhiro-t:
Agreed that it's too complicated.. better not to add it for now. Thanks!

PR #218, opened 2015-04-29
update h2load to give connect time and ttfb stats

ericcarlschwartz:
I was doing some SPDY & HTTP2 performance testing w/ h2load and wanted to also get connect time and time to first byte statistics, so I added them. 

Opening a PR in case this is code that others might want. Let me know what you think of my implementation or if you would like me to make any changes.

### DIFF
--- a/src/h2load.cc
+++ b/src/h2load.cc
@@ -150,7 +150,7 @@ void readcb(struct ev_loop *loop, ev_io *w, int revents) {
 
 Client::Client(Worker *worker, size_t req_todo)
     : worker(worker), ssl(nullptr), next_addr(config.addrs), reqidx(0),
-      state(CLIENT_IDLE), req_todo(req_todo), req_started(0), req_done(0),
+      state(CLIENT_IDLE), first_byte_received(false), req_todo(req_todo), req_started(0), req_done(0),
       fd(-1) {
   ev_io_init(&wev, writecb, 0, EV_WRITE);
   ev_io_init(&rev, readcb, 0, EV_READ);
@@ -165,6 +165,8 @@ int Client::do_read() { return readfn(*this); }
 int Client::do_write() { return writefn(*this); }
 
 int Client::connect() {
+  record_start_time(&worker->stats);
+
   while (next_addr) {
     auto addr = next_addr;
     next_addr = next_addr->ai_next;
@@ -469,6 +471,8 @@ int Client::connection_made() {
 
   session->on_connect();
 
+  record_connect_time(&worker->stats);
+  
   auto nreq =
       std::min(req_todo - req_started, (size_t)config.max_concurrent_streams);
 
@@ -519,6 +523,11 @@ int Client::read_clear() {
     if (on_read(buf, nread) != 0) {
       return -1;
     }
+   
+    if (!first_byte_received) {
+      first_byte_received = true;
+      record_time_to_first_byte(&worker->stats);
+    }     
   }
 
   return 0;
@@ -641,6 +650,11 @@ int Client::read_tls() {
     if (on_read(buf, rv) != 0) {
       return -1;
     }
+ 
+    if (!first_byte_received) {
+      first_byte_received = true;
+      record_time_to_first_byte(&worker->stats);
+    }    
   }
 }
 
@@ -691,6 +705,18 @@ void Client::record_request_time(RequestStat *req_stat) {
   req_stat->request_time = std::chrono::steady_clock::now();
 }
 
+void Client::record_start_time(Stats *stat) {
+  stat->start_times.push_back(std::chrono::steady_clock::now());
+}
+
+void Client::record_connect_time(Stats *stat) {
+  stat->connect_times.push_back(std::chrono::steady_clock::now());
+}
+
+void Client::record_time_to_first_byte(Stats *stat) {
+  stat->time_to_first_bytes.push_back(std::chrono::steady_clock::now());
+}
+
 void Client::signal_write() { ev_io_start(worker->loop, &wev); }
 
 Worker::Worker(uint32_t id, SSL_CTX *ssl_ctx, size_t req_todo, size_t nclients,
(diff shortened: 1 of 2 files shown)


tatsuhiro-t:
Thank you for this awesome PR. I'm reviewing right now, and will have some comments.

src/h2load.cc
```diff
@@ -165,6 +165,9 @@ int Client::do_read() { return readfn(*this); }
 int Client::do_write() { return writefn(*this); }
 
 int Client::connect() {
+  auto tmp_stat = &worker->stats;
+  record_start_time(tmp_stat);
```
tatsuhiro-t: We can pass the address of worker->stats directly to the function:
```
record_start_time(&worker->stats);
```

src/h2load.cc
```diff
@@ -469,6 +472,9 @@ int Client::connection_made() {
 
   session->on_connect();
 
+  auto tmp_stat = &worker->stats;
+  record_connect_time(tmp_stat);
```
tatsuhiro-t: same as above

src/h2load.cc
```diff
@@ -503,7 +509,7 @@ int Client::read_clear() {
 
   for (;;) {
     ssize_t nread;
-    while ((nread = read(fd, buf, sizeof(buf))) == -1 && errno == EINTR)
+    while ((nread = read(fd, buf, sizeof(buf))) == -1 && errno == EINTR)
```
tatsuhiro-t: there are some trailing spaces, clang-config-3.5 can easily removes them.

src/h2load.cc
```diff
@@ -730,20 +758,37 @@ void Worker::run() {
   ev_run(loop, 0);
 }
 
+//TODO make this work with other stats
```
tatsuhiro-t: please remove this line

src/h2load.cc
```diff
@@ -691,6 +707,18 @@ void Client::record_request_time(RequestStat *req_stat) {
...
+}
+
+void Client::record_connect_time(Stats *stat) {
+  stat->connect_time = std::chrono::steady_clock::now();
+}
+
+void Client::record_time_to_first_byte(Stats *stat) {
+  stat->time_to_first_byte = std::chrono::steady_clock::now();
+}
+
```
tatsuhiro-t: I think these recorded period of time is per worker.  But worker can handle multiple Client object which has its own connection to the server.  So probably we should record them per Client not per worker.

src/h2load.cc
```diff
@@ -730,20 +758,37 @@ void Worker::run() {
...
+        case STAT_CONNECT:
+          stat1 = stat.connect_time;
+          stat2 = stat.start_time;
+          break;
+        case STAT_FIRST_BYTE:
+          stat1 = stat.time_to_first_byte;
+          stat2 = stat.start_time;
+          break;
+      }
+      auto t = std::chrono::duration_cast<std::chrono::microseconds>(stat1 - stat2);
```
tatsuhiro-t: w->stats is per worker (I think it should be per client, see above comment), but this code loops though all request.

src/h2load.h
```diff
@@ -132,6 +142,12 @@ struct Stats {
   std::array<size_t, 6> status;
   // The statistics per request
   std::vector<RequestStat> req_stats;
+   // time connect starts
+  std::chrono::steady_clock::time_point start_time;
+  // time to connect
+  std::chrono::steady_clock::time_point connect_time;
+  // time to first byte
+  std::chrono::steady_clock::time_point time_to_first_byte;
```
tatsuhiro-t: I think these should be per Client not Worker.

tatsuhiro-t:
I made several comments.
The biggest concern is currently newly added time_points are per worker.  h2load Worker handles multiple Clients, and each Client makes a connection to the server.  So if I understand correctly, these figures are per connection, therefore should be sampled per Client, not per Worker.

ericcarlschwartz:
@tatsuhiro-t ah very cool. thanks for the clarification there. i was under the impression that each client had its own worker but if that's wrong i can definitely refactor and fix this up! are the workers returned to a pool at some point? how exactly does the sharing happen?

ericcarlschwartz:
ah i looked through the code and understand better what you meant now, sorry. since all of the timing calculations seem to rely on working with the workers, i'm thinking i'm going to add the capability for each worker to hold a list of these times for all of its clients?

src/h2load.cc
ericcarlschwartz: note to myself to use clang to fix formatting

src/h2load.cc
ericcarlschwartz: fixed.

src/h2load.cc
ericcarlschwartz: fixed.

src/h2load.cc
ericcarlschwartz: fixed.

src/h2load.cc
ericcarlschwartz: ah yeah. i think it might make sense for me to make a separate within_sd for STAT_CONNECT and STAT_FIRST_BYTE cases.

src/h2load.cc
ericcarlschwartz: i now have each worker keeping track of all the times for its clients

src/h2load.h
ericcarlschwartz: i now have each worker keeping track of all the times for its clients

ericcarlschwartz:
I updated per your comments, let me know what you think! (also if the formatting is still of please let me know)

ericcarlschwartz:
also if you would like me to squash these commits back down into a single commit let me know

tatsuhiro-t:
Looks good. Thank you!  Please squash commits into one, then I'll merge this into master.

tatsuhiro-t:
Formatting issue is hard to get done right manually, so I'll do clang-format-3.5 to your commit and merge. This is purely editorial.

ericcarlschwartz:
Hey @tatsuhiro-t, I squashed the commits into the first one. Hope other people can get some good use out of this :)

tatsuhiro-t:
Thank you!  Merged now.

PR #221, opened 2015-05-02
h2load_spdy_session errno include

alex-nalivko:
Compilation error around errno when building with Dockerfile.android


tatsuhiro-t:

PR #225, opened 2015-05-04
Add systemd service file

wzyboy:
A service file for systemd distribution (Arch Linux, Ubuntu 15.04+, etc), much simpler than SysV init scripts. :-)


tatsuhiro-t:
Cool!  Thank you!

PR #227, opened 2015-05-06
Add Upstart configuration file

wzyboy:
Upstart is used in Ubuntu 6.10 -> Ubuntu 14.10 as the default init daemon.

Starting with Ubuntu 15.04, Ubuntu switched to systemd.

Ubuntu Server LTS releases (12.04 and 14.04) are still using Upstart as their default init daemon.

This is the Upstart configuration file for those releases. :-)


tatsuhiro-t:
Thank you!  Merged now.

PR #232, opened 2015-05-14
Access violation in buffers

etcimon:
When adding a large amount of data that spans to multiple chunks, the pointer is incremented by the wrong value.

### DIFF
--- a/lib/nghttp2_buf.c
+++ b/lib/nghttp2_buf.c
@@ -320,7 +320,7 @@ int nghttp2_bufs_add(nghttp2_bufs *bufs, const void *data, size_t len) {
     }
 
     buf->last = nghttp2_cpymem(buf->last, p, nwrite);
-    p += len;
+    p += nwrite;
     len -= nwrite;
   }
 


tatsuhiro-t:
Thank you!  This is really nasty bug.

etcimon:
> Thank you! This is really nasty bug.

My pleasure. Btw, do you know if deferred data frames should be added back into ob_da_pq for the next send event? I can't seem to find how `nghttp2_stream_defer_item` would queue the item again

tatsuhiro-t:
There are 2 cases when `nghttp2_stream_defer_item` is called:

First case is by stream level flow control (`NGHTTP2_STREAM_FLAG_DEFERRED_FLOW_CONTROL`).  In this case, it will be resumed by `nghttp2_stream_resume_deferred_item` when WINDOW_UPDATE is received.
It internally checks dependency tree, and if it is eligible to send, then it is queued.

Second case is requested by user returning `NGHTTP2_ERR_DEFERRED` from `nghttp2_data_source_read_callback`.  In this case, application has to call `nghttp2_resume_data` with that stream ID, then it calls `nghttp2_stream_resume_deferred_item`, just like flow control case.

etcimon:
> application has to call nghttp2_resume_data with that stream ID, then it calls nghttp2_stream_resume_deferred_item, just like flow control case.

Thank you, that's what I was looking for.

PR #236, opened 2015-05-15
fix Value stored to 'rv' is never read found by Clang Analyzer

alagoutte:


tatsuhiro-t:

PR #242, opened 2015-05-25
Catch and log failure to set TCP_NODELAY

fwiesel:
Setting TCP_NODELAY may fail. Instead of failing the routine, log it and continue.
It seems, that setting TCP_NODELAY in python on Mac OS X is not necessarily supported.


tatsuhiro-t:
Thank you!  Merged now.

PR #246, opened 2015-05-31
Expose client certificate, if available

fwiesel:
Client certificates can use used for authentication/authorization in the server,
so expose them similar to the remote address.


tatsuhiro-t:
Thank you!  Merged now.

PR #250, opened 2015-06-04
configure: fix bashism

vapier:
The == operator is not in POSIX; use = like normal.

Reported-by: Mikael Magnusson <mikachu@gmail.com>


tatsuhiro-t:
Thank you!  Merged now.

PR #257, opened 2015-06-17
enable third-party for asio_lib too

vapier:
Since this library uses the third-party subdir, make sure we auto-enable
it when that lib is turned on.


tatsuhiro-t:
Thank you!  Merged now.

PR #271, opened 2015-06-27
Add comment on HAR on pushed objects

acesso:
It seems that HAR 1.2 does not have a proper representaion for pushed objects. I could´t just add a custom parameter so I think a comment at response will not hurt anybody.

### DIFF
--- a/src/nghttp.cc
+++ b/src/nghttp.cc
@@ -1443,6 +1443,10 @@ void HttpClient::output_har(FILE *outfile) {
     json_object_set_new(response, "headersSize", json_integer(-1));
     json_object_set_new(response, "bodySize", json_integer(-1));
 
+    auto pushed = req->stream_id % 2 == 0;
+
+    json_object_set_new(response, "comment", json_string(pushed ? "Pushed Object" : ""));
+
     json_object_set_new(entry, "cache", json_object());
 
     auto timings = json_object();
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Thank you!  I'm ok to add comment to response.  I review the HAR spec, it shows that entry (element of entries array) and request also can have comment.  Do you consider to add comment to one of those, instead of response?  It might be more logical to add comment under entry object since it covers both request and response, but I'm open to this decision.

acesso:
As I can see a PUSH_PROMISE may not be used by a request in case that server sends an object that will not be claimed, at least chrome accept the stream and does not claim after that, sounds leaky but I will do some more tests to confirm this. 

is there any other chance that a pushed stream may not be represented in HAR?

tatsuhiro-t:
I'm not aware of it.  HAR is just client side logging of request and response.  HAR specification does not know "push" concept, so probably we have to deal with it using some hack.

acesso:
Sorry for the mess.

1 - Agreed that entry object its the better place for comments to show that it was pushed.
2 - forget about the chrome thing, it is related to a completely different issue, will check this in the proper place later.

3 - Since we are adding comments to entries array could we also add the "connection" and use stream ID as value?  In the spec 1.2 recomends the connection ID or unique ID, but I think the stream ID may be more consistent there.

I´ve tried to do this way but req->stream_id does not cast that way (it needs to be string)

    json_object_set_new(entry, "connection", json_string(req->stream_id.c_str()));

Some minor hacks on the 1.2 like those should help us to show the real needs to the next spec of HAR

tatsuhiro-t:
To make std::string from unsigned integer, use util::utos().  stream_id is signed integer, but good stream has always positive value, so it should work.

FYI, see the interesting proposal from @summerwind about extension to include HTTP/2 framing information to HAR #275.  Its proposal is written in Japanese, google translation will do the job.

tatsuhiro-t:
LGTM.  Merged now.  Thank you!

PR #294, opened 2015-07-13
Adding a grand total time counter

acesso:
Maybe not the best place for this, but since we are on statistics I think this could be helpfull

### DIFF
--- a/src/nghttp.cc
+++ b/src/nghttp.cc
@@ -1930,6 +1930,8 @@ sorted by 'complete'
 id  responseEnd requestStart  process code size request path)" << std::endl;
 
   const auto &base = client.timing.connect_end_time;
+  const auto &start = client.timing.start_time;
+  auto total = std::chrono::duration_cast<std::chrono::microseconds>(base - start);
   for (const auto &req : reqs) {
     auto response_end = std::chrono::duration_cast<std::chrono::microseconds>(
         req->timing.response_end_time - base);
@@ -1948,6 +1950,7 @@ id  responseEnd requestStart  process code size request path)" << std::endl;
               << util::utos_with_unit(req->response_len) << " "
               << req->make_reqpath() << std::endl;
   }
+  std::cout << "Total: " << util::format_duration(total) << std::endl;
 }
 } // namespace
 


tatsuhiro-t:
Thank you for PR.
client.timing.connect_end_time is time point when connection was established, not the session ended.
So "total" is actually the duration between the time when session start and the time when connection was established.
Probably, we need to measure the time point when session ended, for example, in https://github.com/tatsuhiro-t/nghttp2/blob/fb7775e382ef62189dfe98e93474fcefa116e0b4/src/nghttp.cc#L2087
and take difference against session start time.

acesso:
Thanks for your input.

What I´m trying to check is the push gain by running several requests to a same resource with and without push. h2load is great for stress tests but not for ploting distinct requests.

For instance https://www.4site.com.br/acesso/index.html (content with PUSH) and https://www.4site.com.br/acesso/index2.html with no push (or I could use --no-push), avg delay between boxes is 90ms with no loss (could do that with different delays after that with netem but this is actually the real network delay).

I´ve added a record_end_time at the sugested line with same type of start_time (std::chrono::steady_clock::time_point)m considering that all requests are done in ev_run(loop, 0)
used a "total" by the difference against domain lookup end, as follow:

    auto total = std::chrono::duration_cast<std::chrono::microseconds>(client.timing.end_time - client.timing.domain_lookup_end_time);

the results I got so far https://www.4site.com.br/acesso/push.png , still not conclusive.

Maybe there are too little objects or they are too small in consideration with window-bits size to notice a great gain using PUSH.

I know this maybe not part of master branch and is not explicited on TR.

tatsuhiro-t:
For without push, it seems there are plenty of time for client to issue sub resources and server prepare's their responses, while server is sending main HTML (stream id=13), so there is not much difference in this particular case.  Not sure what is the optimal situation for best push experience.  But HTTP/2 also optimizes concurrent requests, so it depends on the time the client discovers the links in HTML.  Slower it gets, push would be faster.  Huge latency also probably favors push.

acesso:
Hello,

I´ve notived now that taking values directly from nghttp client maybe be not necessary, just parsing har file seems to do the trick and already have all timmings. It will also be simpler to compare it with har files from other browsers.

I´ll continue looking for the best situation for push experience using nghttpx on server side.

Thanks

PR #299, opened 2015-07-15
h2load add rate and num connections options

nshoemaker:
I would like to add the ability to specify a constant test time for h2load. In order to specify a constant time, I have added two new options: one to specify the rate at which connections are made, and another to specify the total number of connections to be made throughout the test. 

Feature 1 -- connection rate
------
The new rate option for h2load will allow users to specify the number of connections made per second.

-r, --rate=N
Specifies the fixed rate at which connections are created. The rate must be a positive integer, representing the number of connections to be made per second. When the rate is 0, the program will run as it normally does, creating connections at whatever variable rate it wants. The default value for this option is 0. To use the -r option, the -T option and either the -C or -n option must be specified as well. 

Feature 2 -- number of connections
-----
The new num_conns option of h2load will allow users to specify the total number of connections created throughout the test. 

-C, --num-conns=N
Specifies the total number of connections to create. The total number of connections must be a positive integer. On each connection, -m requests are made. The test stops as soon as the N connections have either completed or failed. When the number of connections is 0, the program will run as it normally does and create as many connections as it needs in order to make the -n requests specified. The default value for this option is 0. The -n option is not required if the -C option is being used. The -C option will be ignored if the -r option is not being used.

Usage
-----
With these new options, there are now two ways to set a constant test time:

1. # of seconds = floor( -C / -r )
2. # of seconds = floor( -n / ( -r * -m ) )

Either method can be used individually. If used together, the test will stop at the smaller of the two times specified. 

Examples:
Time Method 1:
```
h2load -r2 -C4 https://www.google.com
```
This test will run for (-C / -r) = (4 / 2) = 2 seconds, making 2 connections per second, with one request per connection (because only one uri was specified). 

Time Method 2:
```
h2load -r2 -n16 https://www.google.com https://www.google.com
```
This test will run for ( -n / ( -r * -m ) ) = ( 16 / ( 2 * 2 ) ) = 4 seconds, making 2 connections per second, each with two requests to make a total of 4 requests per second. 

Both Methods:
```
h2load -r2 -n16 -C4 https://www.google.com https://www.google.com
-C, -n: warning: two different test times are being set. 
The smaller of the two will be chosen and the test will run for 2 seconds.
starting benchmark...
Protocol: TLSv1.2

finished in 2.26s, 3.53256 req/s, 183.91KB/s
requests: 8 total, 8 started, 8 done, 8 succeeded, 0 failed, 0 errored
status codes: 8 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 426479 bytes total, 3139 bytes headers, 422648 bytes data
time for request:   113.72ms    150.84ms    131.61ms     16.86ms    62.50%
time for connect:   108.47ms    121.60ms    115.32ms      4.65ms    50.00%
time to 1st byte:   108.62ms    121.76ms    115.49ms      4.66ms    50.00%
```
This test has conflicting times being set, so a warning pops up, notifying the user of which time is being used. 

In this case, the -n16 option will be ignored and the test will run for ( -C / -r ) = ( 4 / 2 ) = 2 seconds, making two connections per second, each with two requests to make a total of 8 requests.

### DIFF
--- a/src/h2load.cc
+++ b/src/h2load.cc
@@ -74,8 +74,8 @@ namespace h2load {
 Config::Config()
     : data_length(-1), addrs(nullptr), nreqs(1), nclients(1), nthreads(1),
       max_concurrent_streams(-1), window_bits(30), connection_window_bits(30),
-      no_tls_proto(PROTO_HTTP2), data_fd(-1), port(0), default_port(0),
-      verbose(false), nconns(0), rate(0), current_worker(0) {}
+      rate(0), nconns(0), no_tls_proto(PROTO_HTTP2), data_fd(-1), port(0), default_port(0),
+      verbose(false), current_worker(0) {}
 
 Config::~Config() {
   freeaddrinfo(addrs);
--- a/src/h2load.cc
+++ b/src/h2load.cc
@@ -1083,7 +1083,7 @@ Options:
               completed or failed. When the number of connections is
               0, the program will run as it normally does, creating as
               many connections as it needs in order to make the '-n'
-              requests specified. The defauly value for this option is
+              requests specified. The default value for this option is
               0. The '-n' option is not required if the '-C' option
               is being used.
   -v, --verbose
--- a/src/h2load.cc
+++ b/src/h2load.cc
@@ -75,7 +75,8 @@ Config::Config()
     : data_length(-1), addrs(nullptr), nreqs(1), nclients(1), nthreads(1),
       max_concurrent_streams(-1), window_bits(30), connection_window_bits(30),
       no_tls_proto(PROTO_HTTP2), data_fd(-1), port(0), default_port(0),
-      verbose(false) {}
+      verbose(false), 
+      nconns(0), rate(0), current_worker(0) {}
 
 Config::~Config() {
   freeaddrinfo(addrs);
@@ -83,6 +84,10 @@ Config::~Config() {
   if (data_fd != -1) {
     close(data_fd);
   }
+}
+
+bool Config::is_rate_mode() {    
+  return (this->rate != 0);    
 }
 
 Config config;
@@ -261,7 +266,7 @@ void Client::process_abandoned_streams() {
 }
 
 void Client::report_progress() {
-  if (worker->id == 0 &&
+  if (!worker->config->is_rate_mode() && worker->id == 0 &&
       worker->stats.req_done % worker->progress_interval == 0) {
     std::cout << "progress: "
               << worker->stats.req_done * 100 / worker->stats.req_todo
@@ -963,6 +968,33 @@ std::vector<std::string> read_uri_from_file(std::istream &infile) {
 } // namespace
 
 namespace {
+// Called every second when rate mode is being used
+void second_timeout_cb(EV_P_ ev_timer *w, int revents) {
+  auto config = static_cast<Config *>(w->data);
+
+  auto nclients_per_worker = config->rate;
+  auto nreqs_per_worker = config->max_concurrent_streams * config->rate;
+
+  if(config->current_worker >= std::max(0. , (config->seconds - 1.))) {
+    nclients_per_worker = config->rate + config->conns_remainder;
+    nreqs_per_worker = (int)config->max_concurrent_streams *
+                       (config->rate + config->conns_remainder);
+    ev_timer_stop(config->rate_loop, w);
+  }
+
+  config->workers.push_back(make_unique<Worker>(config->current_worker,
+                                                config->ssl_ctx,
+                                                nreqs_per_worker,
+                                                nclients_per_worker,
+                                                config));
+
+  config->current_worker++;
+
+  config->workers.back()->run();
+}
+} // namespace
+
+namespace {
 void print_version(std::ostream &out) {
   out << "h2load nghttp2/" NGHTTP2_VERSION << std::endl;
 }
@@ -1027,6 +1059,7 @@ Options:
   -p, --no-tls-proto=<PROTOID>
               Specify ALPN identifier of the  protocol to be used when
               accessing http URI without SSL/TLS.)";
+
 #ifdef HAVE_SPDYLAY
   out << R"(
               Available protocols: spdy/2, spdy/3, spdy/3.1 and )";
@@ -1039,6 +1072,24 @@ Options:
   -d, --data=<FILE>
               Post FILE to  server.  The request method  is changed to
               POST.
+  -r, --rate=<N>
+              Specified the fixed rate at which connections are
+              created. The rate must be a positive integer,
+              representing the number of connections to be made per
+              second. When the rate is 0, the program will run as it
+              normally does, creating connections at whatever variable
+              rate it wants. The default value for this option is 0.
+  -C, --num-conns=<N>
+              Specifies the total number of connections to create. The
+              total number of connections must be a positive integer.
+              On each connection, '-m' requests are made. The test
+              stops once as soon as the N connections have either
+              completed or failed. When the number of connections is
+              0, the program will run as it normally does, creating as
+              many connections as it needs in order to make the '-n'
+              requests specified. The defauly value for this option is
+              0. The '-n' option is not required if the '-C' option
+              is being used.
   -v, --verbose
               Output debug information.
   --version   Display version information and exit.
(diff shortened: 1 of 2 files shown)


tatsuhiro-t:
Thank you for PR.  This functionality is very interesting.  I'll review the code.

src/h2load.cc
```diff
@@ -75,7 +75,7 @@ Config::Config()
     : data_length(-1), addrs(nullptr), nreqs(1), nclients(1), nthreads(1),
       max_concurrent_streams(-1), window_bits(30), connection_window_bits(30),
       no_tls_proto(PROTO_HTTP2), data_fd(-1), port(0), default_port(0),
-      verbose(false) {}
+      verbose(false), nconns(0), rate(0), current_worker(0) {}
```
tatsuhiro-t: The fields in initializer list must be ordered as they appear in struct.  I got following warnings:
```
h2load.cc:78:7: warning: field 'verbose' will be initialized after field 'nconns'
      [-Wreorder]
      verbose(false), nconns(0), rate(0), current_worker(0) {}
      ^
h2load.cc:78:23: warning: field 'nconns' will be initialized after field 'rate'
      [-Wreorder]
      verbose(false), nconns(0), rate(0), current_worker(0) {}
                      ^
```

src/h2load.cc
```diff
@@ -963,6 +965,31 @@ std::vector<std::string> read_uri_from_file(std::istream &infile) {
...
 
 namespace {
+// Called every second when rate mode is being used
+void second_timeout_cb(EV_P_ ev_timer *w, int revents) {
+  auto config = static_cast<Config *>(w->data);
+
+  auto nclients_per_worker = config->rate;
+  auto nreqs_per_worker = config->max_concurrent_streams * config->rate;
+
+  if (config->current_worker >= std::max(0., (config->seconds - 1.))) {
```
tatsuhiro-t: Not sure why you use double value here, since config->current_worker and config->seconds are both integers (ssize_t).

src/h2load.cc
```diff
@@ -1035,14 +1063,34 @@ void print_help(std::ostream &out) {
...
+              rate it wants. The default value for this option is 0.
+  -C, --num-conns=<N>
+              Specifies the total number of connections to create. The
+              total number of connections must be a positive integer.
+              On each connection, '-m' requests are made. The test
+              stops once as soon as the N connections have either
+              completed or failed. When the number of connections is
+              0, the program will run as it normally does, creating as
+              many connections as it needs in order to make the '-n'
+              requests specified. The defauly value for this option is
```
tatsuhiro-t: s/defauly/default/

src/h2load.cc
```diff
@@ -1405,46 +1519,90 @@ int main(int argc, char **argv) {
...
-  }
+    for (auto &fut : futures) {
+      fut.get();
+    }
 #endif // NOTHREADS
+  }    //! config.is_rate_mode()
+  // if in rate mode, create a new worker each second
+  else {
+    // set various config values
+    config.seconds = std::min(n_time, c_time);
```
tatsuhiro-t: Probably, this should be moved to L1567?

src/h2load.cc
```diff
@@ -1405,46 +1519,90 @@ int main(int argc, char **argv) {
...
+
+    if ((int)config.nreqs < config.nconns) {
+      config.seconds = c_time;
+      config.workers.reserve(config.seconds);
```
tatsuhiro-t: It would be better to reserve space just one, after config.seconds are determined after these if else.

src/h2load.cc
```diff
@@ -1405,46 +1519,90 @@ int main(int argc, char **argv) {
...
+    struct ev_loop *rate_loop = EV_DEFAULT;
+
+    config.rate_loop = rate_loop;
+
+    // giving the second_timeout_cb access to config
+    timeout_watcher.data = &config;
+
+    ev_init(&timeout_watcher, second_timeout_cb);
+    timeout_watcher.repeat = 1.;
+    ev_timer_again(rate_loop, &timeout_watcher);
```
tatsuhiro-t: Not sure this is intentional, but current implementation wastes first 1 second, since timeout fires after 1 second.  Probably, we need to first manually call second_timeout_cb here?

tatsuhiro-t:
In this patch, we use one worker thread per connection.  Have you considered the processing mode to let one worker thread to handle multiple connections as we do in normal mode?  Perhaps, we can calculate how many clients per thread (-t option) before hand, and add Client to thread until number of client per thread fills up.

src/h2load.cc
nshoemaker: ccfe7f8

src/h2load.cc
nshoemaker: Now comparing as ssize_t: ce00c50

src/h2load.cc
nshoemaker: 7d2ea42

src/h2load.cc
nshoemaker: Moved: eeba121

src/h2load.cc
nshoemaker: eeba121

src/h2load.cc
nshoemaker: This was not intentional originally, but when I implemented it the way you suggest, the times are being under-reported (rather than over-reported as they are currently).

src/h2load.cc
```diff
@@ -1238,6 +1304,23 @@ int main(int argc, char **argv) {
...
+
+  if (config.rate < 0) {
+    std::cerr << "-r: the rate at which connections are made "
+              << "cannot be negative." << std::endl;
+    exit(EXIT_FAILURE);
+  }
+
+  if (config.rate != 0 && config.nthreads != 1) {
+    std::cerr << "-r, -t: warning: the -t option will be ignored when the -r "
+              << "option is in use." << std::endl;
```
nshoemaker: As a first attempt, I decided to ignore the -t option. I do agree that adding something similar to the -t option would be nice. Maybe this is something I could consider for a second patch?

shukitchan:
IMHO, I think we can leverage threads in this way. Assume # of threads is t

1) pre-calculate the # of connections per second per thread = r / t
2) pre-calculate the # of connections per thread = C / t
3) Add a callback for the worker that will be run every second when in rate mode
4) The callback will create r / t connections till the total # of connections reach C / t 
5)  Also i think rps should be total # of requests divided by (C / r) instead of by the actual

src/h2load.cc
```diff
@@ -963,6 +965,31 @@ std::vector<std::string> read_uri_from_file(std::istream &infile) {
...
+void second_timeout_cb(EV_P_ ev_timer *w, int revents) {
+  auto config = static_cast<Config *>(w->data);
+
+  auto nclients_per_worker = config->rate;
+  auto nreqs_per_worker = config->max_concurrent_streams * config->rate;
+
+  if (config->current_worker >= std::max((ssize_t)0, (config->seconds - 1))) {
+    nclients_per_worker = config->rate + config->conns_remainder;
+    nreqs_per_worker = (int)config->max_concurrent_streams *
+                       (config->rate + config->conns_remainder);
```
tatsuhiro-t: I think we should assign 0 to config->conns_remainder here, without that, we keep adding remainder to all subsequent worker, which increases connections and requests.

src/h2load.cc
tatsuhiro-t: If I understand that "under reported" means finished time becomes 4.0s instead of 5.0s, for example, with "-r2 -C10", I think it is fine, the program works as intended.
Perhaps, we should change the phrase "the test will run for N seconds" to something else, since it is not exactly the current program performs.  For slower server, it may take more time to complete requests.  Perhaps, we should just stick to the connection rate?

tatsuhiro-t:
> 1) pre-calculate the # of connections per second per thread = r / t
> 2) pre-calculate the # of connections per thread = C / t
> 3) Add a callback for the worker that will be run every second when in rate mode
> 4) The callback will create r / t connections till the total # of connections reach C / t 

Yeah, something like that.  I'm fine to left these threading handling to the future patch.

> 5) Also i think rps should be total # of requests divided by (C / r) instead of by the actual duration

I'm not sure this is the way to go.  (C / r) does not represent the time the requests are processed.  The slower servers may take more time than (C / r).

src/h2load.cc
tatsuhiro-t: > As a first attempt, I decided to ignore the -t option. I do agree that adding something similar to the -t option would be nice.

I'm fine wit that.  Current patch is clean enough to add -t support later easily.

src/h2load.cc
nshoemaker: In this paradigm, # of workers == # of seconds, so the remainders are only added to the last worker.

src/h2load.cc
nshoemaker: That's a good point. We could print out something more definite, like the total number of requests that are going to be made. I made that change here: c1c177a

src/h2load.cc
tatsuhiro-t: you are right, I completely missed "if" statement above.

nshoemaker:
I think I've addressed all of the comments you brought up. I agree that the multi threading stuff can be saved for a second patch. Let me know if there is anything else you would like me to fix.

tatsuhiro-t:
Thank you.  I will merge this PR tomorrow.

tatsuhiro-t:
Thank you.  Merged now.

PR #302, opened 2015-07-17
nghttpx: Add AES-256-CBC encryption for TLS session ticket

tatsuhiro-t:
This PR adds AES-256-CBC cipher to encrypt TLS session ticket.
For this purpose, we added --tls-ticket-cipher option.  The available values are aes-128-cbc (this is default), and aes-256-cbc.
Since we used the same file format with nginx and apache, we used 16 bytes key for HMAC SHA-256.
It seems this is not a big problem today, but we have not figured out the reason why 16 bytes are used.
So this PR attempts to fix this.  When --tls-ticket-key-file is not used, we now use 32 bytes for HMAC key,as suggested in RFC 5077.  When --tls-ticket-key-file is used, and aes-128-cbc is given in new --tls-ticket-cipher option, 16 bytes HMAC key is used for backward compatibility.  On the other hand, with file and aes-256-cbc in --tls-ticket-cipher, full 32 bytes HMAC key is used.

### DIFF
--- a/gennghttpxfun.py
+++ b/gennghttpxfun.py
@@ -91,6 +91,7 @@ OPTIONS = [
     "header-field-buffer",
     "max-header-fields",
     "include",
+    "tls-ticket-cipher",
     "conf",
 ]
 
--- a/src/shrpx-unittest.cc
+++ b/src/shrpx-unittest.cc
@@ -122,6 +122,8 @@ int main(int argc, char *argv[]) {
                    shrpx::test_shrpx_config_parse_log_format) ||
       !CU_add_test(pSuite, "config_read_tls_ticket_key_file",
                    shrpx::test_shrpx_config_read_tls_ticket_key_file) ||
+      !CU_add_test(pSuite, "config_read_tls_ticket_key_file_aes_256",
+                   shrpx::test_shrpx_config_read_tls_ticket_key_file_aes_256) ||
       !CU_add_test(pSuite, "config_match_downstream_addr_group",
                    shrpx::test_shrpx_config_match_downstream_addr_group) ||
       !CU_add_test(pSuite, "util_streq", shrpx::test_util_streq) ||
--- a/src/shrpx.cc
+++ b/src/shrpx.cc
@@ -627,8 +627,21 @@ void renew_ticket_key_cb(struct ev_loop *loop, ev_timer *w, int revents) {
     ticket_keys->keys.resize(1);
   }
 
-  if (RAND_bytes(reinterpret_cast<unsigned char *>(&ticket_keys->keys[0]),
-                 sizeof(ticket_keys->keys[0])) == 0) {
+  auto &new_key = ticket_keys->keys[0];
+  new_key.cipher = get_config()->tls_ticket_cipher;
+  new_key.hmac = EVP_sha256();
+  new_key.hmac_keylen = EVP_MD_size(new_key.hmac);
+
+  assert(EVP_CIPHER_key_length(new_key.cipher) <= sizeof(new_key.data.enc_key));
+  assert(new_key.hmac_keylen <= sizeof(new_key.data.hmac_key));
+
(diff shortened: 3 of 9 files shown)


64616E69656C:
I'd recommend CTR / CCM, GCM oder SIV in favour of CBC. CBC is considered obsolete, though not a severe security issue. Matthew Green has a good post about this:
http://blog.cryptographyengineering.com/2013/02/why-i-hate-cbc-mac.html

tatsuhiro-t:
I can add CTR.  As far as I know, current OpenSSL interface does not accept CCM and GCM cipher for session ticket encryption.  It is just designed based on recommended ticket construction described in RFC 5077, only taking into account AES128 CBC + SHA256.

64616E69656C:
My thoughts about using GCM/CCM/CTR in favour of CBC where,
 - CBC is protected from bit flipping when HMAC-SHA2 is applied after ticket encryption (although HMAC is vulnerable to timed side channel attacks, afaik), 
 - *BUT* AES-CBC-HMAC-SHA2 would result in a small overhead in package size, which might have slight impacts for mobile users (thinking about TLS TTFB)
 - *AND* using AES-<stream cipher>, HMAC could be avoided (but it would of course require integrity checks).

Not sure about specific CCM/GCM usage in session ticket encryption though, it's there in libressl, maybe openssl has this only in aead branch.

tatsuhiro-t:
This specific to session ticket encryption and its OpenSSL implementation.   The cipher itself is supported in OpenSSL.  And session ticket itself does not mandate specific cipher suite.  The problem is OpenSSL's session ticket callback.  It does not offer the interface for the application to pass the additional parameters required for CCM/GCM.
I think libressl uses same ticket callback, since nghttpx can be compiled with it.
If openssl is "fixed" to accept GCM, then it would be better to use just AES-256-GCM.

64616E69656C:
Got it. Think this will be possible in 1.1.0, if I read this correct: https://github.com/openssl/openssl/blob/master/CHANGES#L388-L404

tatsuhiro-t:
Thank you.  These new functions are very interesting.  It seems there is no change in ticket callback, will need to check the implementation...

64616E69656C:
I'm just thinking out loud here, and since I'm in the stage where I'm more learning to understand how the code works rather than writing it myself just yet, I gotta ask: Wouldn't it be possible to include NaCl as a third party and staticly link tls session ticket encryption against nacl? This could gain the possibility to use salsa20 based ciphers for this purpose.

etcimon:
Botan has had AES-256/GCM session tickets for quite some time now: https://github.com/randombit/botan/blob/master/src/lib/tls/tls_session.cpp#L158

tatsuhiro-t:
Botan has done the good job regarding session ticket encryption.

If I understand correctly, you mean using salsa20 from statically linked NaCl and still use openssl infrastructure, right?  Then unfortunately, it does not work, since OpenSSL session ticket encryption requires ciphers implemented in OpenSSL (specifically, EVP_* function).  The thing is that encryption itself is done by OpenSSL library.  The application can only select a cipher from the ones OpenSSL supports.  So we cannot encrypt ticket using cipher from external source.

64616E69656C:
Well, more like dropping openssl for ssl ticket session encryption completly. Not sure if that's possible (let openssl handle TLS connections (encryption, authentication and key exchange) and use another lib just for session ticket encryption). If it is, NaCl was just a thought that came to mind, but I agree with @etcimon that botan would/could be a good choice for this, code quality looks amazing and it provides a good range of state-of-the-art encryption.

tatsuhiro-t:
Current openssl implementation cannot do this.
I feel it is a bit too much work to replace openssl with botan just for ticket encryption cipher.
I'm not expert of security, but it seems even aes-256-cbc for ticket encryption purpose is good enough today, and now the key is renewed every 1 hour and ticket life time is 12 hours.

PR #305, opened 2015-07-21
Fix rm: cannot remove ‘*.rst’: No such file or directory when "make c…

alagoutte:
Fix rm: cannot remove ‘*.rst’: No such file or directory when "make clean"

When you don't generated documentation via make html


tatsuhiro-t:
Thank you!  Merged now.

PR #309, opened 2015-07-24
[h2load] Adding -t option support to rate mode

nshoemaker:
-t option now supported in rate mode. Instead of generating new Workers each second, now we have a fixed set of Workers that generate new Clients each second.

Some new specifications: 
-t <= -r (connection rate must be greater than or equal to the number of threads used.)
-t <= -C (total number of connections made must be greater than or equal to the number of threads used.)


tatsuhiro-t:
Awesome!  Merged now.

PR #322, opened 2015-08-10
h2load timeout options

nshoemaker:
It would be nice if h2load could timeout certain connections and close them, rather than waiting indefinitely until they close by themselves. In order to do this, I would like to add two new options to h2load: one to specify a way to timeout a connection after a fixed period of time, and another to timeout a connection if there has been no activity on said connection for a fixed period of time. 

Option 1 -- active timeout
========
This new option for h2load will ensure that all connections have a max time they are allowed to be open.

-T, --connection-active-timeout=N
Specifies the maximum time that h2load is willing to keep a connection open, regardless of the activity on said connection. N must be a positive integer, specifying the number of seconds to wait. When no timeout value is set (either active or inactive), h2load will keep a connection open indefinitely, waiting for a response.
 
After the last request is sent on any given connection, the connection will wait --connection-active-timeout seconds before it is closed by h2load. Any timed-out requests will be added to the timeout count.

Option 2 -- inactivity timeout
=======
This new option for h2load allows connections to be timed out after a certain amount of inactivity.

-N, --connection-inactivity-timeout=<N>
Specifies the amount of time N that h2load is willing to wait to see activity on a given connection. N must be a positive integer, specifying the number of seconds to wait. When no timeout value is set (either active or inactive), h2load will keep a connection open indefinitely, waiting for a response.
 
If a connection sees no activity within --connection-inactivity-timeout seconds, it will be closed by h2load. Any timed-out requests will be added to the timeout count.

Usage
=======
```
h2load -r2 -C20 -N2 -T5 https://www.google.com
starting benchmark...
spawning thread #0: 20 concurrent clients, 20 total requests
Protocol: TLSv1.2

finished in 10.11s, 1.87921 req/s, 98.44KB/s
requests: 20 total, 20 started, 20 done, 19 succeeded, 1 failed, 1 errored
status codes: 19 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 1019146 bytes total, 10319 bytes headers, 1006669 bytes data
time for request:    51.98ms     70.07ms     58.85ms      4.83ms    68.42%
time for connect:    39.40ms     66.75ms     45.64ms      6.72ms    85.00%
time to 1st byte:    39.53ms     66.85ms     45.74ms      6.73ms    85.00%
requests timed out:1
```
In this example, one request is timed out due to inactivity. 

The timed-out requests will affect all stats that they are able to completely finish--meaning their time for connect, status codes, etc. will be factored into the stats, but time for request will not as the request did not complete fully before it was timed out. 

Additionally, if the -T or -N option is used, the number of requests timed out will be printed out on a separate line. The number of requests timed out is equal to the number of requests actively timed out or timed out due to inactivity during the test. Timed out requests will count as both failed and errored. 

Have been working on docs for these new options on a separate branch: 01cfc1260c1fb4ec795655d43e52218cbc3472d9 will create separate PR to follow.


tatsuhiro-t:
Thank you!  I'll review this PR.  Next release, v1.2.1 is for stability fix for nghttpx, and this new feature will be merged after that.

tatsuhiro-t:
For documentation, they are generated by `h2load -h` output by makemanpages script.
We do this periodically, so you don't have to do that in PR.

src/h2load.cc
```diff
@@ -187,6 +224,17 @@ Client::Client(Worker *worker, size_t req_todo)
...
+  if (worker->config->conn_inactivity_timeout > 0) {
+    conn_inactivity_watcher.data = this;
+    ev_init(&conn_inactivity_watcher, conn_inactivity_timeout_cb);
+    conn_inactivity_watcher.repeat = worker->config->conn_inactivity_timeout;
+  }
+  if (worker->config->conn_active_timeout > 0) {
+    conn_active_watcher.data = this;
+    ev_timer_init(&conn_active_watcher, conn_active_timeout_cb,
+                  worker->config->conn_active_timeout, 0);
+  }
```
tatsuhiro-t: It would be simpler to just initialize both timers regard less of configuration.
It is harmless unless they are actually started.

src/h2load.cc
```diff
@@ -178,6 +180,41 @@ void second_timeout_w_cb(struct ev_loop *loop, ev_timer *w, int revents) {
...
+  if (client->worker->config->conn_inactivity_timeout > 0 &&
+      ev_is_active(&client->conn_inactivity_watcher)) {
+    ev_timer_stop(client->worker->loop, &client->conn_inactivity_watcher);
+  }
+
+  if (util::check_socket_connected(client->fd)) {
+    client->timeout();
+  }
+}
+} // namespace
```
tatsuhiro-t: These 2 functions are essentially same, except for stopping individual timers.
As I wrote below, if we are going to initialize always these timers, it is safe to just stop both here without checking it is activated.  Then, we can achieve the same result using just one function.

src/h2load.cc
```diff
@@ -271,9 +345,32 @@ void Client::disconnect() {
 }
 
 void Client::submit_request() {
+  restart_timeout();
```
tatsuhiro-t: So we need to re-arm timer on every network activity.  Then, it is simpler to do this in writecb and readcb above, instead of calling restart_timeout() in every single hook functions.

src/h2load.cc
```diff
@@ -251,6 +315,16 @@ void Client::fail() {
...
 void Client::disconnect() {
+  if (worker->config->conn_inactivity_timeout > 0 &&
+      ev_is_active(&conn_inactivity_watcher)) {
+    ev_timer_stop(worker->loop, &conn_inactivity_watcher);
+  }
+
+  if (worker->config->conn_active_timeout > 0 &&
+      ev_is_active(&conn_active_watcher)) {
+    ev_timer_stop(worker->loop, &conn_active_watcher);
+  }
```
tatsuhiro-t: Just stop timers without checking its activation status.  See above.

src/h2load.cc
```diff
@@ -271,9 +345,32 @@ void Client::disconnect() {
...
+
+  // if an active timeout is set and this is the last request to be submitted
+  // on this connection, start the active timeout.
+  if (worker->config->conn_active_timeout > 0 && req_started >= req_todo) {
+    ev_timer_start(worker->loop, &conn_active_watcher);
+  }
+}
+
+void Client::process_timedout_streams() {
+  for (auto req_stat : worker->stats.req_stats) {
```
tatsuhiro-t: `&` is required for `req_stat`, otherwise it is a copy of each element in worker->stats.req_stats.
So, do like so:
```
for (auto &req_stat : worker->stats.req_stats) {
```

src/h2load.cc
```diff
@@ -271,9 +345,32 @@ void Client::disconnect() {
...
+    }
+  }
+  auto req_timed_out = req_todo - req_done;
+
+  worker->stats.req_failed += req_timed_out;
+  worker->stats.req_error += req_timed_out;
+  worker->stats.req_done += req_timed_out;
+  worker->stats.req_timedout += req_timed_out;
+
+  req_done = req_todo;
```
tatsuhiro-t: Please use :process_abandoned_streams() to update worker->stats except for req_timeout.

src/h2load.cc
```diff
@@ -1543,17 +1686,16 @@ int main(int argc, char **argv) {
...
     config.nthreads = 1;
   }
-  ssize_t seconds = 0;
 
   if (config.is_rate_mode()) {
 
     // set various config values
     if (config.nreqs < config.nconns) {
-      seconds = c_time;
+      config.seconds = c_time;
```
tatsuhiro-t: I'm curious why seconds is now data field in config object.  What is the purpose for this change?

src/h2load.cc
```diff
@@ -1709,6 +1852,9 @@ time for request: )" << std::setw(10) << util::format_duration(ts.request.min)
             << util::format_duration(ts.ttfb.sd) << std::setw(9)
             << util::dtos(ts.ttfb.within_sd) << "%" << std::endl;
 
+  if (config.conn_inactivity_timeout > 0 || config.conn_active_timeout > 0) {
+    std::cout << R"(requests timed out: )" << stats.req_timedout << std::endl;
+  }
```
tatsuhiro-t: Rather than showing this in separate line, it would be more streamlined to print them after "requests: ..., N errored" line (e.g., "requests: ..., N errored, M timeout"), since this is a subset of failure, just like "errored."  What do you think about this?

tatsuhiro-t:
For documentation, we have additional file to insert after `h2load -h` output, which is doc/h2load.h2r.
Depending on the outcome of the review, we need to add "timeout" to OUTPUT section.

src/h2load.cc
nshoemaker: I think that the timeout stat fits in with the other request stats nicely. 

Would you want it to always print out, or only print out if a timeout was set?

src/h2load.cc
nshoemaker: Consolidated into one callback `conn_timeout_cb()` in 4e5b607.

src/h2load.cc
tatsuhiro-t: Either will do.  Personally, just printing it always is simpler at the moment.

nshoemaker:
I've added timeout to the req stats line and updated the output docs. Please let me know if there is anything else you'd like me to do. Thanks!

src/h2load.cc
```diff
@@ -187,6 +207,14 @@ Client::Client(Worker *worker, size_t req_todo)
 
   wev.data = this;
   rev.data = this;
+
+  conn_inactivity_watcher.data = this;
+  ev_init(&conn_inactivity_watcher, conn_timeout_cb);
+  conn_inactivity_watcher.repeat = worker->config->conn_inactivity_timeout;
```
tatsuhiro-t: We can set repeat in ev_timer_init as well:
```
ev_timer_init(&conn_inactivity_watcher, conn_timeout_cb, 0., worker->config->conn_inactivity_timeout);
```

tatsuhiro-t:
Thank you!  Almost perfect.  I found one cosmetic thing.  After you resolve it, please squash commits into one.  Then I'll pull this PR.

nshoemaker:
No problem! Continuing in #331.

PR #324, opened 2015-08-11
h2load allow -r < -n when -n is not specified

nshoemaker:
With dd97b53554a48b5a1a591dc1094341d5dfd5eb15, the division by zero problem was fixed, but now it forces -n >= -r, even when -n is not specified. 

For instance:
```
h2load -r2 -C2 https://www.google.com
-r, -n: the connection rate must be smaller than or equal to the number of requests.
```
We are seeing this behavior because the default value of -n = 1 so -r > -n in this case.

From the [docs](https://nghttp2.org/documentation/h2load.1.html#cmdoption-h2load-C):
```
 "The -n option is not required if the -C option is being used."
```
If we want to keep -n optional when -C is being used, my change ensures that we only check if -r > -n when -n is specified from the command line. 

This way we still fix the divide by zero problem:
```
h2load -r2 -n1 https://www.google.com 
-r, -n: the connection rate must be smaller than or equal to the number of requests.
```

but also allow -n to be optional when -C is being used:
```
h2load -r2 -C2 https://www.google.com
starting benchmark...
spawning thread #0: 2 concurrent clients, 2 total requests
Protocol: TLSv1.2

finished in 213.05ms, 9.38747 req/s, 491.39KB/s
requests: 2 total, 2 started, 2 done, 2 succeeded, 0 failed, 0 errored
status codes: 2 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 107204 bytes total, 1088 bytes headers, 105894 bytes data
time for request:   124.15ms    126.44ms    125.30ms      1.14ms    50.00%
time for connect:    82.85ms     86.02ms     84.43ms      1.59ms   100.00%
time to 1st byte:    82.99ms     86.10ms     84.54ms      1.55ms    50.00%
```


tatsuhiro-t:
Yeah, you are right, I missed that use case.

I applied your patch, but still gets div by zero if I use -r only:
```
$ h2load -r4 https://localhost/
```

Perhaps, we should also check -C is used or not?

nshoemaker:
Yes, I noticed that as well. I added a check to make sure -r is accompanied by -n or -C.

```
$ h2load -r2 https://localhost/
-r: the rate option must be used with either the -n option or the -C option.
```

tatsuhiro-t:
Looks good.  Could you rebase and squash commit into one commit? Then I'll pull this PR.

nshoemaker:
Continuing with #329.

Issue #325, opened 2015-08-11
Don't use virtual /dev/stderr


thinred:
Hi,
I'm trying to run nghttp2x under systemd and I get this kind of warnings/errors:

    nghttpx[20201]: Failed to open errorlog file /dev/stderr

Without --errorlog-syslog this message is shown 3 times and nghttp2x does not start.
With --errorlog-syslog nghttp2x works, but the message is shown 2 times nevertheless.

It seems that systemd does not like /dev/stderr:
http://lists.freedesktop.org/archives/systemd-devel/2015-April/031218.html

I propose to replace the logic of "/dev/stderr" in the source code with something that duplicates the stderr under fd = 2 (and using /proc/self/fd/2 does not count).

Cheers!
Tomasz

tatsuhiro-t:
Fixed by PR #326

PR #326, opened 2015-08-11
Map /dev/stdout and /dev/stderr to their descriptors.

thinred:
This seems to do the trick for the issue #325.

It may also be used to simplify the Android-specific workarounds with /proc/self/df/2.

### DIFF
--- a/src/shrpx.cc
+++ b/src/shrpx.cc
@@ -1059,12 +1059,7 @@ void fill_default_config() {
   mod_config()->accesslog_file = nullptr;
   mod_config()->accesslog_syslog = false;
   mod_config()->accesslog_format = parse_log_format(DEFAULT_ACCESSLOG_FORMAT);
-#if defined(__ANDROID__) || defined(ANDROID)
-  // Android does not have /dev/stderr.  Use /proc/self/fd/2 instead.
-  mod_config()->errorlog_file = strcopy("/proc/self/fd/2");
-#else  // !__ANDROID__ && ANDROID
   mod_config()->errorlog_file = strcopy("/dev/stderr");
-#endif // !__ANDROID__ && ANDROID
   mod_config()->errorlog_syslog = false;
   mod_config()->conf_path = strcopy("/etc/nghttpx/nghttpx.conf");
   mod_config()->syslog_facility = LOG_DAEMON;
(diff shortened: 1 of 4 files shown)


tatsuhiro-t:
Thank you for PR!
We need to set CLOEXEC flag to new file descriptor from dup.
Since it is a bit difficult to make this happen atomically, perhaps we can treat /dev/stderr and /dev/stdout specially in reopen_log_files in shrpx_log.cc so that we just use STDERR_FILENO and STDOUT_FILENO without opening additional file descriptors.

thinred:
Oh, right, there are some forks here and there in the code. Actually I wanted to do as you say, but then I figured out that instead of reopening the file I can simply "reopen" the descriptor. I seemed clean and simpler to me. We can use dup3() to set O_CLOEXEC automatically, too. :)

tatsuhiro-t:
We have already some #if-def code to support platform which lacks O_CLOEXEC flag.  For that platform, descriptor may leak depending on the severe timing of fork.  So it is ok to use O_CLOEXEC to dup3 and provide replacement code for other platforms.  That said, in portability point of view, it is better not to use dup3 if we can avoid it...

thinred:
I wonder whether there are platforms that have O_CLOEXEC, but no dup3(). Well, anyway, I'll cook something without it. Stay tuned.

thinred:
Hi again,
please review this. As a bonus, a patch on top of this removes some android-specific code which is now obsolete.

thinred:
There is still a small issue with the code (and it existed even before, actually).
In redirect_stderr_to_errorlog, the stderr fd is replaced with errorlog_fd and so stderr is closed. Now, if we execv the process with updated configuration it *may* happen that the errorlog is from a regular file to stderr once again, but this won't work since stderr was already closed.
I believe that a solution is to dup stderr early in the program and use it instead.

thinred:
I've pushed some more code, I haven't tested the last commit, because I'd like to know what you think.

tatsuhiro-t:
Thank you.  I completely missed this issue.  I agree with your idea for the fix.  It seems it is working fine.

tatsuhiro-t:
@thinred Do you think these commits are ready to merge?

thinred:
They solve the problem for me and I'm fairly confident that they don't break anything.
I'd say yes.

tatsuhiro-t:
OK, thank you.  I also tested this and worked fine for me.  Merged now.

PR #329, opened 2015-08-13
h2load allow -r < -n when -n is not specified 

nshoemaker:
Continuing #324 as single commit.


tatsuhiro-t:
Thank you!  Merged now.

PR #330, opened 2015-08-14
Add Timing-script and base URI support

LPardue:
Hi Tatsuhurio,

This pull request adds two features to h2load:

1) A new timing script file to allow h2load to issue requests according to a script. A copy of the script is executed by all clients. The file format is tab-separated, field one defines the time to execute the request, measured from the start in milliseconds to microsecond resolution. Field two defines the URI. A new command-line parameter is added to h2load to define the use of this file. The number of requests is contrained to the length of the file, it doesn't seem logical to loop back to start of script.

Example file format
```
    100.0	https://localhost/index.html
    200.0	/style.css
    300	/script.js
```

2) A base URI can be defined in a new command-line parameter. This is used to define the scheme, host and port number to be used for all requests. This overrides all other behaviour. This feature allows for simplification of input files, removing the details and allowing for reuse of the files for different hosts or protocols.

There are some rough edges that may need to be worked on but I am keen to get this infront of you for review.

Please let me know what you think.

### DIFF
--- a/src/h2load.cc
+++ b/src/h2load.cc
@@ -76,7 +76,7 @@ Config::Config()
       max_concurrent_streams(-1), window_bits(30), connection_window_bits(30),
       rate(0), nconns(0), conn_active_timeout(0), conn_inactivity_timeout(0),
       no_tls_proto(PROTO_HTTP2), data_fd(-1), port(0), default_port(0),
-      verbose(false) {}
+      verbose(false), timing_script(false) {}
 
 Config::~Config() {
   freeaddrinfo(addrs);
@@ -87,7 +87,7 @@ Config::~Config() {
 }
 
 bool Config::is_rate_mode() const { return (this->rate != 0); }
-
+bool Config::has_base_uri() const { return (!this->base_uri.empty()); }
 Config config;
 
 namespace {
@@ -196,6 +196,51 @@ void conn_timeout_cb(EV_P_ ev_timer *w, int revents) {
     client->timeout();
   }
 }
+}//namespace
+
+namespace {
+bool check_stop_client_request_timeout(Client *client, ev_timer *w) {
+  auto nreq = client->req_todo - client->req_started;
+
+  if (nreq == 0 ||
+      client->streams.size() >= (size_t)config.max_concurrent_streams) {
+    // no more requests to make, stop timer
+    ev_timer_stop(client->worker->loop, w);
+    return true;
+  }
+
+  return false;
+}
+} // namespace
+
+namespace {
+void client_request_timeout_cb(struct ev_loop *loop, ev_timer *w, int revents) {
+  auto client = static_cast<Client *>(w->data);
+
+  client->submit_request();
+  client->signal_write();
+
+  if (check_stop_client_request_timeout(client, w)) {
+    return;
+  }
+
+  ev_tstamp duration =
+      config.timings[client->reqidx] - config.timings[client->reqidx - 1];
+
+  while (duration < 1e-9) {
+    client->submit_request();
+    client->signal_write();
+    if (check_stop_client_request_timeout(client, w)) {
+      return;
+    }
+
+    duration =
+        config.timings[client->reqidx] - config.timings[client->reqidx - 1];
+  }
+
+  client->request_timeout_watcher.repeat = duration;
+  ev_timer_again(client->worker->loop, &client->request_timeout_watcher);
+}
 } // namespace
 
 Client::Client(Worker *worker, size_t req_todo)
@@ -215,9 +260,15 @@ Client::Client(Worker *worker, size_t req_todo)
   ev_timer_init(&conn_active_watcher, conn_timeout_cb,
                 worker->config->conn_active_timeout, 0.);
   conn_active_watcher.data = this;
+
+  ev_timer_init(&request_timeout_watcher, client_request_timeout_cb, 0., 0.);
+  request_timeout_watcher.data = this;
 }
 
-Client::~Client() { disconnect(); }
+Client::~Client() {
+  ev_timer_stop(worker->loop, &request_timeout_watcher);
+  disconnect();
+}
 
 int Client::do_read() { return readfn(*this); }
 int Client::do_write() { return writefn(*this); }
@@ -473,9 +524,11 @@ void Client::on_stream_close(int32_t stream_id, bool success,
     return;
   }
 
-  if (req_started < req_todo) {
-    submit_request();
-    return;
+  if (!config.timing_script) {
+    if (req_started < req_todo) {
+      submit_request();
+      return;
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


tatsuhiro-t:
Thank you!  This is interesting feature.  Will review and leave comments.

src/h2load.cc
```diff
@@ -481,13 +529,26 @@ int Client::connection_made() {
...
+
+    ev_tstamp duration = config.timings[reqidx];
 
+    while (duration < 0.0001) {
+      submit_request();
+      duration = config.timings[reqidx];
+    }
+    request_timeout_watcher.data = this;
+    ev_init(&request_timeout_watcher, client_request_timeout_cb);
+    request_timeout_watcher.repeat = duration;
```
tatsuhiro-t: I thin it is better to initialize request_timeout_watcher in Client constructor, which eliminates uninitialized member field warning with static analyzer, and we can unconditionally "ev_timer_stop" it in destructor.  Please use "ev_timer_init" instead of ev_init generic macro.

src/h2load.cc
```diff
@@ -1003,6 +1070,44 @@ std::vector<std::string> read_uri_from_file(std::istream &infile) {
...
+    if (pos == std::string::npos) {
+      std::cerr << "Invalid line format detected, no tab character at line "
+                << line_count << ". \n\t" << script_line << std::endl;
+      exit(EXIT_FAILURE);
+    }
+
+    try {
+      auto v = std::stod(script_line.substr(0, pos)) / 1000.0;
+      timings.push_back(v);
+      uris.push_back(script_line.substr(pos + 1, script_line.length()));
```
tatsuhiro-t: We conventionally use std::string::size instead of length

src/h2load.cc
```diff
@@ -1003,6 +1070,44 @@ std::vector<std::string> read_uri_from_file(std::istream &infile) {
...
+
+    std::size_t pos = script_line.find("\t");
+    if (pos == std::string::npos) {
+      std::cerr << "Invalid line format detected, no tab character at line "
+                << line_count << ". \n\t" << script_line << std::endl;
+      exit(EXIT_FAILURE);
+    }
+
+    try {
+      auto v = std::stod(script_line.substr(0, pos)) / 1000.0;
```
tatsuhiro-t: We conventionally don't use error handling using exception.  Could you use strtod here?

src/h2load.cc
```diff
@@ -1249,6 +1377,13 @@ int main(int argc, char **argv) {
         exit(EXIT_FAILURE);
       }
       break;
+    case 'B':
+      if (!parse_base_uri(optarg)) {
+        std::cerr << "invalid base URI: " << optarg << std::endl;
+        exit(EXIT_FAILURE);
+      }
+      config.base_uri = std::string(optarg);
```
tatsuhiro-t: You can assign optarg directly to config.base_uri.

src/h2load.cc
```diff
@@ -1268,6 +1403,10 @@ int main(int argc, char **argv) {
         // ciphers option
         config.ciphers = optarg;
         break;
+      case 3:
+        config.ifile = std::string(optarg);
```
tatsuhiro-t: See above comment

tatsuhiro-t:
It may be a bit annoying to correctly set -n option to match the number of entries in timing file.
One possible workaround is ignore -n command-line, and assign the number of entries in timing file to config.nreqs.  Of course, if we have use case to use -n in conjunction to --timing-script-file, this idea not fly..

src/h2load.cc
```diff
@@ -178,6 +178,52 @@ void second_timeout_w_cb(struct ev_loop *loop, ev_timer *w, int revents) {
...
 
+namespace {
+bool check_cancel_client_request_timeout(Client *client) {
+  auto nreq = client->req_todo - client->req_started;
+
+  if (nreq == 0 ||
+      client->streams.size() >= (size_t)config.max_concurrent_streams) {
+    // no more requests to make, let timer expire
+    client->request_timeout_watcher.repeat = 0.;
+    ev_timer_again(client->worker->loop, &client->request_timeout_watcher);
```
tatsuhiro-t: I think we can just stop timer here without touching repeat.

src/h2load.cc
```diff
@@ -178,6 +178,52 @@ void second_timeout_w_cb(struct ev_loop *loop, ev_timer *w, int revents) {
...
+  client->signal_write();
+
+  if (check_cancel_client_request_timeout(client)) {
+    return;
+  }
+
+  ev_tstamp duration =
+      config.timings[client->reqidx] - config.timings[client->reqidx - 1];
+
+  while (duration < 0.0001) {
```
tatsuhiro-t: Usually, we use 1e-9 or 1e-6 to express small delta.

src/h2load.cc
```diff
@@ -1493,6 +1647,7 @@ int main(int argc, char **argv) {
     shared_nva.emplace_back(":authority", config.host);
   }
   shared_nva.emplace_back(":method", config.data_fd == -1 ? "GET" : "POST");
+  shared_nva.emplace_back("user-agent", "h2load nghttp2/" NGHTTP2_VERSION);
```
LPardue: I overlooked explicitly mentioning this change in the initial pull request. Is this acceptable? There could be room to make this more configurable (command-line parameter etc) if you want me to expand it.

Adding the user agent field with a default value has helped in my testing/analysis, especially public-facing servers where various other user agents (also with no UA string) probe the server.

LPardue:
Thanks for the feedback! I'll address the comments and prepare an updated request.

src/h2load.cc
tatsuhiro-t: I agree with you for the use case, especially for the UA string analysis on public server.  nghttp2.org sometimes gets lots of requests seemingly generated by h2load.

Theoretically, adding header will increase server load (HPACK huffman decoding), so it could reduce some numbers.  But I don't think it is not much an issue, since we can re-run h2load as needed.

LPardue:
> It may be a bit annoying to correctly set -n option to match the number of entries in timing file.
One possible workaround is ignore -n command-line, and assign the number of entries in timing file to config.nreqs. Of course, if we have use case to use -n in conjunction to --timing-script-file, this idea not fly..

I agree, my initial intention was to cover all cases but I missed the one where requests is not defined (and defaults to 1). I will modify my value checking so that:

1) if -n is not given with timing script, `config.nreqs = uris.size();`
2) if -n is given with timing script, `config.nreq` must be less than or equal to `uris.size()` otherwise we clamp to `uris.size()`
3) if -n is given with timing script and is less than `uris.size()` then the first n-requests of the script are executed. This is handy for constraining a load test, for instance in development testing.

I think this covers the cases without ignoring -n completely.

LPardue:
Commits added to address the comments. I can squash these down once we are happy with things.

tatsuhiro-t:
Thank you.  The patch looks good to me.
Unfortunately, there are conflict due to merge of #331.  Could you resolve the conflict?

src/h2load.cc
```diff
@@ -481,13 +534,25 @@ int Client::connection_made() {
...
-    submit_request();
-  }
+    for (; nreq > 0; --nreq) {
+      submit_request();
+    }
+  } else {
+
+    ev_tstamp duration = config.timings[reqidx];
 
+    while (duration < 0.0001) {
```
tatsuhiro-t: We could use 1e-9 here as above; I missed this in previous review

LPardue:
cd47582 resolves merge conflicts and uses `(duration < 1e-9)`.

tatsuhiro-t:
Thank you for quick work!  Merged now.

LPardue:
Thanks for merging, this lets us try out some cool things with the latest version of h2load.

Apologies for the leftover conflict markers, it took a few attempts and I missed out on that!

tatsuhiro-t:
No problem.  It is always hard to find those inside string literal since it does not cause compile error..

PR #331, opened 2015-08-18
h2load timeout options

nshoemaker:
Continuing #322. 

It would be nice if h2load could timeout certain connections and close them, rather than waiting indefinitely until they close by themselves. In order to do this, I would like to add two new options to h2load: one to specify a way to timeout a connection after a fixed period of time, and another to timeout a connection if there has been no activity on said connection for a fixed period of time. 

Option 1 -- active timeout
========
This new option for h2load will ensure that all connections have a max time they are allowed to be open.

-T, --connection-active-timeout=N
Specifies the maximum time that h2load is willing to keep a connection open, regardless of the activity on said connection. N must be a positive integer, specifying the number of seconds to wait. When no timeout value is set (either active or inactive), h2load will keep a connection open indefinitely, waiting for a response.
 
After the last request is sent on any given connection, the connection will wait --connection-active-timeout seconds before it is closed by h2load. Any timed-out requests will be added to the timeout count.

Option 2 -- inactivity timeout
=======
This new option for h2load allows connections to be timed out after a certain amount of inactivity.

-N, --connection-inactivity-timeout=<N>
Specifies the amount of time N that h2load is willing to wait to see activity on a given connection. N must be a positive integer, specifying the number of seconds to wait. When no timeout value is set (either active or inactive), h2load will keep a connection open indefinitely, waiting for a response.
 
If a connection sees no activity within --connection-inactivity-timeout seconds, it will be closed by h2load. Any timed-out requests will be added to the timeout count.

Usage
=======
```
h2load -r2 -C20 -N2 -T5 https://www.google.com
starting benchmark...
spawning thread #0: 20 concurrent clients, 20 total requests
Protocol: TLSv1.2

finished in 10.11s, 1.87921 req/s, 98.44KB/s
requests: 20 total, 20 started, 20 done, 19 succeeded, 1 failed, 1 errored
status codes: 19 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 1019146 bytes total, 10319 bytes headers, 1006669 bytes data
time for request:    51.98ms     70.07ms     58.85ms      4.83ms    68.42%
time for connect:    39.40ms     66.75ms     45.64ms      6.72ms    85.00%
time to 1st byte:    39.53ms     66.85ms     45.74ms      6.73ms    85.00%
requests timed out:1
```
In this example, one request is timed out due to inactivity. 

The timed-out requests will affect all stats that they are able to completely finish--meaning their time for connect, status codes, etc. will be factored into the stats, but time for request will not as the request did not complete fully before it was timed out. 

Additionally, if the -T or -N option is used, the number of requests timed out will be printed out on a separate line. The number of requests timed out is equal to the number of requests actively timed out or timed out due to inactivity during the test. Timed out requests will count as both failed and errored.

### DIFF
--- a/doc/h2load.h2r
+++ b/doc/h2load.h2r
@@ -19,6 +19,9 @@ requests
     This is the subset of the number reported in ``failed`` and most
     likely the network level failures or stream was reset by
     RST_STREAM.
+  timeout
+    The number of requests whose connection timed out before they 
+    were completed.
 
 status codes
   The number of status code h2load received.
--- a/src/h2load.cc
+++ b/src/h2load.cc
@@ -74,8 +74,9 @@ namespace h2load {
 Config::Config()
     : data_length(-1), addrs(nullptr), nreqs(1), nclients(1), nthreads(1),
       max_concurrent_streams(-1), window_bits(30), connection_window_bits(30),
-      rate(0), nconns(0), no_tls_proto(PROTO_HTTP2), data_fd(-1), port(0),
-      default_port(0), verbose(false) {}
+      rate(0), nconns(0), conn_active_timeout(0), conn_inactivity_timeout(0),
+      no_tls_proto(PROTO_HTTP2), data_fd(-1), port(0), default_port(0), 
+      verbose(false) {}
 
 Config::~Config() {
   freeaddrinfo(addrs);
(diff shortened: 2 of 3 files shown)


tatsuhiro-t:
Thank you!  Merged now.

PR #334, opened 2015-08-19
Fix 332

### LINKED ISSUE #332: Timing script combined with -c
Timing script added capability for number of requests to be assumed from the number of script lines, when `-n` is not defined .

The following example command shows how to trigger a parameter validation bug due to sequencing.

`h2load --timing-script-file='/path/to/file' -c2`

This bug occurs because `config.nreqs` defaults to 1, any value larger than 1 triggers this issue. 

I will look into fixing this. Would it be reasonable move the input file parsing block earlier in the validation sequence?

LPardue:
This attempts to fix #332.

Also now correctly allows each client to execute a full test script, without having to manually state `-n`.

### DIFF
--- a/src/h2load.cc
+++ b/src/h2load.cc
@@ -1527,6 +1527,66 @@ int main(int argc, char **argv) {
     }
   }
 
+  if (config.nclients == 0) {
+    std::cerr << "-c: the number of clients must be strictly greater than 0."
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Thank you!  Merged now.

PR #340, opened 2015-09-01
h2load: Improve checking for timing script imput, prevent false posit…

### LINKED ISSUE #339: h2load: timing script file false postitive error
When loading h2load, certain conditions may lead to errno being set to a nonzero value during startup. These conditions do not prevent the successful loading of h2load, and normal usage should continue. However, in combination with the timing script, we see a false positive error returned for parsing/validation of the timing script e.g. `if (end == start || errno != 0) `.

Better practice would be to set errno to 0 before the call to `strtod`. A more specific check of errno (`ERANGE`) would also help.

LPardue:
Fixes #339

### DIFF
--- a/src/h2load.cc
+++ b/src/h2load.cc
@@ -1164,9 +1164,14 @@ void read_script_from_file(std::istream &infile,
     char *end;
     auto v = std::strtod(start, &end);
 
-    if (end == start || errno != 0) {
+    errno = 0;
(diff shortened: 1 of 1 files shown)


LPardue:
Added some logical range checking for time value to be given to libev. Negative value, NaN or INF are not very helpful...

tatsuhiro-t:

Issue #333, opened 2015-08-19
h2load: Add HTTP/1.1 support 


LPardue:
h2load is a great tool for load testing of HTTP/2 servers, supporting both h2c and h2. SPDY is also supported if build with spdylay library.

This issue is a feature request for the addition of HTTP/1.1 support  (cleartext and encrypted) to h2load.

Having a single tool capable of supporting multiple versions of HTTP will allow for fairer comparisons between the results of load test runs. Other tools such as  [Apache Bench (ab)](http://httpd.apache.org/docs/2.4/programs/ab.html) or [wrk](https://github.com/wg/wrk) support HTTP/1.1 but comparing the results to h2load is unfair, due to different architectures and calculation of statistics.

## Design Goals
* Support for HTTP/1.1, cleartext and encrypted
* Convenient selection of HTTP/1.1 usage:
  * Use existing mechanisms where possible
* Convenient fallback to HTTP/.1:
  * E.g. failure of ALPN will cause fallback to HTTP/1.1 (if enabled)

## Constraints
* h2load architecture uses a single connection per origin per client
  * UA's typically use up to 6-10 HTTP/1.1 connections per origin

## Implementation details
* Add Http1Session class and use it in h2load.cc
* Add new command-line parameter for client ALPN/NPN protocol identifier list, mirror [nghttpx parameter] (https://nghttp2.org/documentation/nghttpx.1.html#cmdoption-nghttpx--npn-list)
* Add support for http/1.1 to [`-p` option](https://nghttp2.org/documentation/h2load.1.html#cmdoption-h2load-p)

tatsuhiro-t:
I think user defined -m option is effectively ignored for HTTP/1, and it is treated as if -m1 is set, and we can reuse connection (keep-alive) to send multiple requests one by one.
For web server which does support HTTP/1 only, if server does not select anything in ALPN, we should fallback to HTTP/1.

LPardue:
Thanks for the feedback. I've incorporated the fallback into Parent post.

> I think user defined -m option is effectively ignored for HTTP/1, and it is treated as if -m1 is set, and we can reuse connection (keep-alive) to send multiple requests one by one.

Are you suggesting that HTTP Pipelining should be effectively disabled?

tatsuhiro-t:
When I was writing the previous post, I thought h1 pipelining should be disabled, so yes.
[...]

PR #351, opened 2015-09-11
Initial HTTP/1.1 capability. Add npn-list option to h2load. Make NPN/…

LPardue:
…ALPN more runtime dependent.

Here is the initial release of my work to add HTTP/1.1 support to h2load. I think there will be some iteration of these changes to address review comments and known issues (below). Once the PR is accepted it should address #333.

### Protocol negotiation and cleartext operation
Many of the changes relate to application protocol negotiation, specifying a list of options at the command-line and then using these to negotiate with the server. To allow code reuse, some functions used elsewhere have been extracted to util. 

h2load had some `debug` functions that seem to be a remnant from older versions, these have been replaced and the program now reports things in a more consistent way.

`http/1.1` has been added to the end of the default list of supported applications for TLS connections. This means that h2load has an ability to fallback to HTTP/1.1 (for example, if the server does not support h2). This could be considered useful/not useful for a user, so the program prints feedback on the negotiation process. The user can overwrite the default list with `--npn-list` option. The user could prevent HTTP/1.1 fallback behavior by supplying a list of h2-only protocol identifiers. The name `--npn-list` has been chosen to match nghttpx naming but can be changed if it is confusing.
 
Previously, a complete failure to negotiate protocol could occur without without a helpful notice to the user (and it would print results stats). The behavior has been changed to explicitly state this case and exit the program with failure condition. 

Cleartext connections use h2c as default, this can be overridden to `http/1.1` using `-p, --no-tls-proto` option. NB. help/docs need updating to list `http/1.1` as an available protocol. No fallback behavior happens in cleartext mode.

### HTTP/1.1
In this PR the HTTP/1.1 capabilities are basic, no pipelining support has been implemented or supported. 

For internal tracking of h2 requests/responses h2load uses stream IDs, managed by nghttp library. We are not so fortunate to have this capability in HTTP/1.1. To preserve the APIs and methodology around streams,  I use a rudimentary 'notional stream ID' for request and responses that are tracked within h2load_http1_session. A better solution may be required going forward.

### Time to 1st byte
time to 1st byte reports when using HTTP/1.1 seem inaccurate. For example, below are the statistics from tests on a server that supports both h2 and HTTP/1.1:

```
http/1.1
time for request:   297.12ms    297.12ms    297.12ms         0us   100.00%
time for connect:     3.53ms      3.53ms      3.53ms         0us   100.00%
time to 1st byte:   113.62ms    113.62ms    113.62ms         0us   100.00%


h2
time for request:   144.97ms    144.97ms    144.97ms         0us   100.00%
time for connect:     4.05ms      4.05ms      4.05ms         0us   100.00%
time to 1st byte:     4.17ms      4.17ms      4.17ms         0us   100.00%
```
Setting of TTFB is done in `client::read_clear` or `client_read_tls`, I suspect the difference in reading HTTP/1.1 data affects this but can't figure out where to fix this.

### Limited testing
Testing has only been performed with a small number of requests for a single resource. I need to do further testing especially with the timing script functionality.


tatsuhiro-t:
Awesome!

> Time to 1st byte

Not much difference observed in my environment between h1 and h2 for normal production build.
If I enabled debug options on (e.g., ASAN), h1 gets much higher TTFB just like you wrote.
In that case, moving the code which record TTFB before on_read() fixes the problem.

tatsuhiro-t:
Wait a sec, it seems cleartext version is ok, but if we do tls and use large number of requests, TTFB for h1 gets higher, and the workaround I mentioned in the previous post does not help.

I think this is because h2 server can send back SETTINGS frame as soon as it accepted connection without waiting from request, which means higher TTFB.  OTOH, h1 server only returns HTTP response after request was processed, so if server is loaded, responses are generally lagged, and resulted in higher TTFB value.

Perhaps, we should define TTFB as first HTTP protocol level.  I mean, instead of recording TTFB in first read from socket, record it when first HTTP response body is received.

tatsuhiro-t:
From wikipedia https://en.wikipedia.org/wiki/Time_To_First_Byte
> Time To First Byte (TTFB) is a measurement used as an indication of the responsiveness of a webserver or other network resource.
>
>TTFB measures the duration from the virtual user making an HTTP request to the first byte of the page being received by the browser. This time is made up of the socket connection time, the time taken to send the HTTP request, and the time taken to get the first byte of the page.

So it probably means first byte of response body, rather than first socket read.  Of course, this is fundamental h2load issue, not in this PR particular

tatsuhiro-t:
With 0d8c8ca, TTFB is now recorded on the arrival of the first byte of response body, rather than first socket read.  This makes HTTP/1, HTTP/2 and SPDY/3.1 TTFB times similar.

tatsuhiro-t:
For HTTP/1 session, ttfb is integrated like so:
```diff
diff --git a/src/h2load_http1_session.cc b/src/h2load_http1_session.cc
index 596cc61..e560db5 100644
--- a/src/h2load_http1_session.cc
+++ b/src/h2load_http1_session.cc
@@ -71,6 +71,7 @@ int htp_msg_completecb(http_parser *htp) {
   auto session = static_cast<Http1Session *>(htp->data);
   auto client = session->get_client();
 
+  client->record_ttfb();
   client->on_stream_close(session->stream_resp_counter_, true,
                           session->req_stats_[session->stream_resp_counter_]);
 
@@ -81,6 +82,17 @@ int htp_msg_completecb(http_parser *htp) {
 } // namespace
 
 namespace {
+int htp_body_cb(http_parser *htp, const char *data, size_t len) {
+  auto session = static_cast<Http1Session *>(htp->data);
+  auto client = session->get_client();
+
+  client->record_ttfb();
+
+  return 0;
+}
+} // namespace
+
+namespace {
 http_parser_settings htp_hooks = {
     htp_msg_begincb,   // http_cb      on_message_begin;
     nullptr,           // http_data_cb on_url;
@@ -88,7 +100,7 @@ http_parser_settings htp_hooks = {
     nullptr,           // http_data_cb on_header_field;
     nullptr,           // http_data_cb on_header_value;
     nullptr,           // http_cb      on_headers_complete;
-    nullptr,           // http_data_cb on_body;
+    htp_body_cb,       // http_data_cb on_body;
     htp_msg_completecb // http_cb      on_message_complete;
 };
 } // namespace
```

LPardue:
Cool I didn't expect to reveal a different bug. Thanks for fixing it, glad we can improve the overall quality of h2load!

I'll incorporate your h1.1 ttfb change into the PR on Monday.

While on the topic of request stats, I can see a potential use case for an export of all request statistics for offline analysis - a bit like the HAR export in nghttp ``-r, --har``. I will have a further think on this.

tatsuhiro-t:
Thank you.
The request statistics offline analysis is very interesting. For HTTP/2, we are not just interested in how fast the server is, but also the quality of service, such as the priorities, or response times etc.

LPardue:
I have mistakenly messed up this branch while attempting to rebase on latest master and I'm not sure how to rectify it. I'm going to suggest a new pull request that squashes the commits once it is ready.

LPardue:
I am testing with a timing script that should allow for multiple requests in flight before responses are made. E.g.

```
100.0	https://nghttpxserver/index.html
100.0	nghttpxserver/index.html
100.0	nghttpxserver/index.html
100.0	nghttpxserver/index.html
100.0	nghttpxserver/index.html
100.0	nghttpxserver/index.html
100.0	nghttpxserver/index.html
```

Looking at the nghttpx logs (at log level INFO) I see a different sequence of reports between HTTP/1.1 and h2. For HTTP/1.1, I see nghttpx log the first request, then the first response, second request and second response and so on. For h2, I see it log the first seven requests, then the seven responses.

Looking at the packet trace for HTTP/1.1 I see all requests sent in the same packet, and then the responses come in one-by-one over the course of many packets. This implies that pieplining is working but I am confused by the way nghttpx is logging, are you able to shed any light on this?

LPardue:
This is superseded by #356

tatsuhiro-t:
Thank you!
As for nghttpx behaviour, basically it processes request one by one, and while processing one request, it does not log next request.  nghttpx continues to read incoming bytes until its read buffer gets full, but it left untouched until current request is finished.
On the other hard, in h2 we can do full multiplex behaviour, so we can process multiple requests concurrently.

LPardue:
Ahh, so from client perspective, HTTP pipelining is being supported but internally nghttpx is doing things serially 1-by-1. This makes sense now.

Closing this PR as we have moved to #356.

PR #356, opened 2015-09-15
Initial HTTP/1.1 capability. Add npn-list option to h2load. Make NPN/…

LPardue:
…ALPN more runtime dependent

The alternate clean history branch that implements HTTP/1.1 as described in #351. Once the PR is accepted it should address #333.

This version addresses the TTFB issue.

I have performed additional testing with the timing script option and all appears to work satisfactorily .

### DIFF
--- a/src/h2load.cc
+++ b/src/h2load.cc
@@ -54,6 +54,7 @@
 
 #include "http-parser/http_parser.h"
 
+#include "h2load_http1_session.h"
 #include "h2load_http2_session.h"
 #ifdef HAVE_SPDYLAY
 #include "h2load_spdy_session.h"
@@ -89,31 +90,6 @@ Config::~Config() {
 bool Config::is_rate_mode() const { return (this->rate != 0); }
 bool Config::has_base_uri() const { return (!this->base_uri.empty()); }
 Config config;
-
-namespace {
-void debug(const char *format, ...) {
-  if (config.verbose) {
-    fprintf(stderr, "[DEBUG] ");
-    va_list ap;
-    va_start(ap, format);
-    vfprintf(stderr, format, ap);
-    va_end(ap);
-  }
-}
-} // namespace
-
-namespace {
-void debug_nextproto_error() {
-#ifdef HAVE_SPDYLAY
-  debug("no supported protocol was negotiated, expected: %s, "
-        "spdy/2, spdy/3, spdy/3.1\n",
-        NGHTTP2_PROTO_VERSION_ID);
-#else  // !HAVE_SPDYLAY
-  debug("no supported protocol was negotiated, expected: %s\n",
-        NGHTTP2_PROTO_VERSION_ID);
-#endif // !HAVE_SPDYLAY
-}
-} // namespace
 
 RequestStat::RequestStat() : data_offset(0), completed(false) {}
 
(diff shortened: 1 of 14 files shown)


tatsuhiro-t:
Thank you very much.  I'll review this PR and post comments if any.

LPardue:
Ok great thanks

src/h2load.cc
```diff
@@ -1507,9 +1571,13 @@ int main(int argc, char **argv) {
         config.ciphers = optarg;
         break;
       case 3:
+        // timing-script option
         config.ifile = optarg;
         config.timing_script = true;
         break;
+      case 4:
+        // npn-list option
+        config.npn_list = util::parse_config_str_list(optarg);
```
tatsuhiro-t: Could you add "break" here?  It would save our lives when we add new option.

src/h2load_http1_session.cc
```diff
@@ -0,0 +1,160 @@
...
+}
+} // namespace
+
+namespace {
+int htp_body_cb(http_parser *htp, const char *data, size_t len) {
+  auto session = static_cast<Http1Session *>(htp->data);
+  auto client = session->get_client();
+
+  client->record_ttfb();
+
```
tatsuhiro-t: I think we need to update response body bytes received here, just like we do in h2, spdy:
```
client->worker->stats.bytes_body += len;
```

src/h2load_http1_session.h
```diff
@@ -0,0 +1,58 @@
...
+  virtual int on_write();
+  virtual void terminate();
+  Client *get_client();
+  int32_t stream_req_counter_;
+  int32_t stream_resp_counter_;
+  std::unordered_map<int32_t, RequestStat *> req_stats_;
+
+private:
+  Client *client_;
+  std::unique_ptr<http_parser> htp_;
```
tatsuhiro-t: I personally embed http_parser struct in Http1Session class since it would avoid additional allocation.
But not sure how much it impacts on h2load performance.  Perhaps, not much.

tatsuhiro-t:
I think we need to mention "http/1.1" in --no-tls-proto option.

src/shrpx_config_test.cc
```diff
@@ -33,31 +33,32 @@
 #include <CUnit/CUnit.h>
 
 #include "shrpx_config.h"
+#include "util.h"
 
 namespace shrpx {
 
 void test_shrpx_config_parse_config_str_list(void) {
```
tatsuhiro-t: Since parse_config_str_list is now defined in util.cc, this test code should be in util_test.cc, renaming as `test_util_parse_config_str_list`

LPardue:
Thanks for comments, I hope to address these all tomorrow!

src/h2load_http1_session.cc
LPardue: I will also add response header bytes via new on_header_field, on_header_value callbacks

tatsuhiro-t: Good catch.  I missed that.

LPardue:
commit ec47dfb should address all comments

tatsuhiro-t:
Thank you!  Merged now.

PR #365, opened 2015-09-23
Do not try to set TCP_NODELAY when frontend is an UNIX socket

rraptorr:
This silences warning log that otherwise spams logs on every accepted
connection.


tatsuhiro-t:
Thank you!  Merged now.

PR #366, opened 2015-09-24
Make traditional init script fail if new config file is broken

rraptorr:
Instead of assuming that upgrade is always successful, check if a new
process actually started. Do not kill old process if the new one
failed.


tatsuhiro-t:
Thank you!  Looks good.  Merged now.

PR #368, opened 2015-09-25
Fix daemon upgrade when running under systemd

rraptorr:
Systemd assumes that service of type simple is stopped when the main
process exits. This causes systemd to kill all nghttpx processes when
doing a process upgrade (via USR2/QUIT signals). Change the service
type to forking which behaves correctly on upgrade.


tatsuhiro-t:
Thank you! Merged now.

PR #375, opened 2015-10-01
add PIDFile directive to systemd service

thinred:
Hi,
since now Type=forking is used, why not use PIDFile= too?

Tomasz

### DIFF
--- a/contrib/nghttpx.service.in
+++ b/contrib/nghttpx.service.in
@@ -5,6 +5,7 @@ After=network.target
 [Service]
 Type=forking
 ExecStart=@bindir@/nghttpx --conf=/etc/nghttpx/nghttpx.conf --pid-file=/run/nghttpx.pid --daemon
+PIDFile=/run/nghttpx.pid
 
 [Install]
 WantedBy=multi-user.target


tatsuhiro-t:
Agreed.  Merged now. Thank you!

rraptorr:
Since I've converted type to forking I'll comment why PIDFile was not used.
If PIDFile is used, systemd assumes that pid file is created and available as soon as parent process exists. Unfortunately that is not the case with nghttpx as the pid file is not created by parent process, like systemd expects, but by child process. This causes a race condition that manifests with errors from systemd about pid file not being available (just stop/start nghttpx several times to reproduce this).

To support PIDFile nghttpx should be converted so that pid file is created by the parent process.

thinred:
Oh my, so this should be reverted, I suppose?
So systemd is able to handle the situation without PIDFile? How does it work?

rraptorr:
I am not that experienced with systemd to say how does it work. With PIDFile it sometimes prints scary looking errors, but in my tests it seemed to work anyway every time. I've chosen no to add PIDFile to avoid people having to worry about those errors.

Personally I'd revert this, but this is not something critical I think.

tatsuhiro-t:
Thank you.  So to use PIDFile correctly, we have to write PID file in parent process after first fork().
Current nghttpx implementation writes PID file in child process (the main daemon process, in systemd manual).  Currently, we use daemon(3) to make nghttpx system daemon, and as long as we use it, we cannot write PID file from parent process since it only returns in child process.
Since it is a bit easier to write daemon call by hand (I did in another project https://github.com/tatsuhiro-t/aria2/blob/4bbc71c8e6c6aa7b5441e30da3a5efc5b810073a/src/daemon.cc#L43), we can import it here, and write pid file in the correct location.  If that is done, can we continue to use PIDFile directive?

tatsuhiro-t:
Hmm, in daemon we do fork() twice.  The process ID we need is the return value of 2nd fork(),  That means the first parent process already gone without writing process ID.

rraptorr:
I am not really sure that all of this is worth it. Without the PIDFile directive all works fine and systemd seems able to detect process pid. It even works after upgrading the process to a new version.
I am not that familiar with systemd to say if you should work on supporting PIDFile. Maybe just remove it for now and see if anyone reports any case that doesn't work?

tatsuhiro-t:
Agreed.  I'll merge your PR. Meanwhile I'll do some research about how PIDFile really works with daemon mode.

wzyboy:
Just traced to here via [the commit history](https://github.com/tatsuhiro-t/nghttp2/commits/master/contrib/nghttpx.service.in) of `contrib/nghttpx.service.in`. This discussion above confused me a bit.

Quoting from `systemd.service(5)`:

> GuessMainPID=
> Takes a boolean value that specifies whether systemd should try to guess the main PID of a service if it cannot be determined reliably. This option is ignored unless Type=forking is set and PIDFile= is unset because for the other types or with an explicitly configured PID file, the main PID is always known. The guessing algorithm might come to incorrect conclusions if a daemon consists of more than one process. If the main PID cannot be determined, failure detection and automatic restarting of a service will not work reliably. Defaults to yes.

So with current setup (`Type=forking` and `PIDFile` unset), systemd would try to guess the main PID. I am not sure how systemd would do that, but it states that the guessing algorithm might come to incorrect conclusions if a daemon consists of more than one process. Nghttpx is a multi-worker damon, so I wonder whether the PID guessing would work reliably...

I could see @rraptorr changed `Type=simple` to `Type=forking` in https://github.com/tatsuhiro-t/nghttp2/commit/1cdd8b1c58f549b45c2a8e8165806b8bb798fe03. If changing the way how PID file get written is problematic, could we consider changing the behavior of the main process to make systemd happy with `Type=simple`? I am not sure...

rraptorr:
With Type=simple you cannot use USR2/QUIT to upgrade running nghttpx as systemd will simply kill everything as soon as the first process exits. That was the whole point of my change.

rraptorr:
From what I've checked systemd simply crawls process tree and searches for a single process that has systemd as a parent. Given how nghttpx works it should be ok (worker process does not have systemd as parent and is ignored).

This might break only in scenarios where executable spawns multiple, separate, daemon processes on startup. I don't think that's common.

wzyboy:
@rraptorr I see. Thanks for explanation!

PR #378, opened 2015-10-02
Added support for RFC 7413 (TCP Fast Open) on nghttpx proxy for listening connections.

serioussam:
This patch enables setting fastopen on listening tcp socket to allow faster connection setup. 
--fastopen option was added to binary which will take limit of number of requests with handshake for security.

More description https://tools.ietf.org/html/rfc7413

### DIFF
--- a/gennghttpxfun.py
+++ b/gennghttpxfun.py
@@ -102,6 +102,7 @@ OPTIONS = [
     "response-phase-file",
     "accept-proxy-protocol",
     "conf",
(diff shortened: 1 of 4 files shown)


tatsuhiro-t:
Thank you!  Please see my inline comment above.

serioussam:
@tatsuhiro-t I have updated the PR with incorporating your comments, Please review

tatsuhiro-t:

Issue #358, opened 2015-09-16
h2load: rate feature modification


LPardue:
Hi,

I am considering an enhancement to the `-r, --rate` feature, where defining a rate between 0 and 1 would allow clients to be created at a rate less than every second. For example, to create a new client every 10 seconds we could provide the rate value `0.1`.

I've done some tinkering in the code to test if this would be possible (and it is by ultimately varying the timeout value given to `timeout_watcher`) but I keep hitting issues with the validation logic, and also calculations when running multiple `--threads`.

I wanted to get a better understanding of the current ruleset for rate mode. Perhaps @nshoemaker may be able to help too? 

For instance, why is this rule `-r, -t: the connection rate must be greater than or equal to the number of threads.` enforced?

Regards
Lucas

LPardue:
Further,

The current rate feature creates N clients after 1 second elapses. This means the creation is not 'smooth' but instead is quantised/discrete on the second boundaries. For large N, the stepping effect of this is going to be more pronounced. 

A solution to this issue would be to allow the definition of the timeout period. This is not directly associated with my first enquiry, but could go towards a more expressive syntax for defining longer timeout periods. Some care would need to be paid to avoid over-complicating the h2load interface but I think it is possible. 

Any thoughts?

tatsuhiro-t:
I think -r > -t rule is done just because making calculation/code easier.  If I allow it, then we get floating point exception.  If we make it robust, perhaps, we reduce -t to -r.

tatsuhiro-t:
Smoothing out for creating connection sounds good.
I'm wondering which is better for load testing scenario, stepping or smooth slope?  Perhaps, we may have both.

LPardue:
> I'm wondering which is better for load testing scenario, stepping or smooth slope? Perhaps, we may have both.

To take it even further, even a smooth linear trend for client connection may not correctly simulate the load on a server. Thinking on more of a macro scale, the rate of connection could be described as a curve over time period. For example, some of the ease in curves described [here](http://easings.net/) and shown below. 

![easing-curves](https://cloud.githubusercontent.com/assets/6571445/9935092/b46c10b6-5d4c-11e5-8771-2e4c1b03546d.png)

This is maybe too much complexity for h2load, but it could be implemented by h2load internally generating a "client creation schedule" based upon rate, and choice of straight line, curves, discrete or smooth.

nshoemaker:
[...]

PR #380, opened 2015-10-05
h2load: Add option for user-definable rate period

LPardue:
This PR relates to #358. The changes take a different approach from my initial suggestion in the issue, its a quick prototype rather than the finished goods but I've done some testing and looks ok so far.

A new option `--rate-period` is added allowing a user to define a period between client creation greater than 1 second (the default is still 1 second so existing functionality is unaffected). This allows a user to define that a total `C` clients should be created at a rate of `rate / rate-period`. 

The effect of combining these parameters may not be immediately obvious to a user, therefore some more information is printed at startup. Most importantly clients may not be concurrent when using rate mode, so we remove that terminology.

Note that this PR does not address the `connection rate must be greater than or equal to the number of threads` constraint. Nor does it address smooth vs. stepped creation.

Lets look at an example, the below command will create 6 clients over 2 threads at a rate of 2 clients per 30 seconds:

```
h2load --npn-list=http/1.1 -phttp/1.1 -Bhttp://192.168.56.101:8080 --threads=2 --rate=2 --rate-period=30 --num-conn=6 /index.html 

starting benchmark...
spawning thread #0: 3 total client(s). Up to 2 client(s) will be created every 30 seconds. 3 total requests
spawning thread #1: 3 total client(s). Up to 2 client(s) will be created every 30 seconds. 3 total requests
Application protocol: http/1.1

finished in 113.85s, x.x req/s, y.yKB/s
requests: 6 total, 6 started, 6 done, 6 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 6 2xx, 0 3xx, 0 4xx, 0 5xx
```

First client on each thread is made as soon as possible, then after 30 seconds two clients are created in each thread. Once all responses complete, h2load finishes.

### DIFF
--- a/src/h2load.cc
+++ b/src/h2load.cc
@@ -74,9 +74,9 @@ namespace h2load {
 Config::Config()
     : data_length(-1), addrs(nullptr), nreqs(1), nclients(1), nthreads(1),
       max_concurrent_streams(-1), window_bits(30), connection_window_bits(30),
-      rate(0), nconns(0), conn_active_timeout(0), conn_inactivity_timeout(0),
-      no_tls_proto(PROTO_HTTP2), data_fd(-1), port(0), default_port(0),
-      verbose(false), timing_script(false) {}
+      rate(0), rate_period(1.0), nconns(0), conn_active_timeout(0),
+      conn_inactivity_timeout(0), no_tls_proto(PROTO_HTTP2), data_fd(-1),
+      port(0), default_port(0), verbose(false), timing_script(false) {}
 
 Config::~Config() {
   freeaddrinfo(addrs);
@@ -136,8 +136,8 @@ void readcb(struct ev_loop *loop, ev_io *w, int revents) {
 } // namespace
 
 namespace {
-// Called every second when rate mode is being used
-void second_timeout_w_cb(struct ev_loop *loop, ev_timer *w, int revents) {
+// Called every rate_period when rate mode is being used
+void rate_period_timeout_w_cb(struct ev_loop *loop, ev_timer *w, int revents) {
   auto worker = static_cast<Worker *>(w->data);
   auto nclients_per_second = worker->rate;
   auto conns_remaining = worker->nclients - worker->nconns_made;
@@ -938,8 +938,9 @@ Worker::Worker(uint32_t id, SSL_CTX *ssl_ctx, size_t req_todo, size_t nclients,
   auto nreqs_per_client = req_todo / nclients;
   auto nreqs_rem = req_todo % nclients;
 
-  // create timer that will go off every second
-  ev_timer_init(&timeout_watcher, second_timeout_w_cb, 0., 1.);
+  // create timer that will go off every rate_period
+  ev_timer_init(&timeout_watcher, rate_period_timeout_w_cb, 0.,
+                config->rate_period);
   timeout_watcher.data = this;
 
   if (!config->is_rate_mode()) {
@@ -974,8 +975,8 @@ void Worker::run() {
   } else {
     ev_timer_again(loop, &timeout_watcher);
 
-    // call callback so that we don't waste the first second
-    second_timeout_w_cb(loop, &timeout_watcher, 0);
+    // call callback so that we don't waste the first rate_period
+    rate_period_timeout_w_cb(loop, &timeout_watcher, 0);
   }
   ev_run(loop, 0);
 }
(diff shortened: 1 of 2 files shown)


LPardue:
Also important to note, I removed the check that disallows rate feature with timing script. In my use case I want to be able to ramp up clients that all execute the same script. Was there a reason that I'm not aware of for this check to have been added?

tatsuhiro-t:
Thank you for PR!
As for the removed check, I could not remember why I did this. It seems to be unnecessary.

tatsuhiro-t:
We can clean up the output such as "spawning ..." in one or 2 places, but it could be done in the subsequent commits after this PR was merged.

LPardue:
> We can clean up the output such as "spawning ..." in one or 2 places, but it could be done in the subsequent commits after this PR was merged.

I'm happy to improve the information more in this PR but wasn't feeling very creative at the time, I'm open to ideas or maybe we do it after PR is merged as you suggest.

src/h2load.cc
```diff
@@ -1610,6 +1619,25 @@ int main(int argc, char **argv) {
...
         config.npn_list = util::parse_config_str_list(optarg);
         break;
+      case 5: {
+        // rate-period
+        const char *start = optarg;
+        char *end;
+        errno = 0;
+        auto v = std::strtod(start, &end);
+
+        if (v < 1.0 || !std::isfinite(v) || end == start || errno != 0) {
```
LPardue: Fixed inconsistency with period value checking, now matches manual

tatsuhiro-t: Thank you!

tatsuhiro-t:
I recalled why I avoided to use both `--timing-script` and rate mode.
The thing is `-c` option is totally ignored when rate mode is used.
In rate mode the number of connections are either `--num-conns` or determined by `-n` and `-r`.
The current code of timing script assumes `-c` clients.  So I just added the check without further thinking about the workaround.
Perhaps, we can just scale number of requests with `--num-conns` or something rather than `-c` option for timing script.  But the code calculating these numbers are already convoluted, so might be better to clean up.

A little bit off topic: In rate mode, it uses `-m` as the number of requests made per connection, which is not the original meaning of `-m` option.
I'm thinking to add new option to specify the number of request made per connection in rate mode, rather than reusing `-m` option in different way.  `-m` should be just the parameter reflecting the protocol detail of HTTP/2 or SPDY.

tatsuhiro-t:
The PR itself looks good to me.  The scaling factor of URIs in timing scripts can be handled in later commits, when we refactor the code around rate mode.
If you agree, I'll merge this PR.

LPardue:
Lets go with the merge!

> Perhaps, we can just scale number of requests with --num-conns or something rather than -c option for timing script.

In my testing the scaling seems to happen already. For example, I have a script file with 7 URIs defined:

```
h2load --threads=2 --timing-script-file='timing.script' -Bhttp://192.168.56.101:8080 --rate=2 --rate-period=1.0 --num-conn=6

-C, -n: warning: number of requests conflict. 
The smaller of the two will be chosen and the test will create 7 total requests.
starting benchmark...
spawning thread #0: 3 total client(s). Up to 2 client(s) will be created every 2 seconds. 21 total requests
spawning thread #1: 3 total client(s). Up to 2 client(s) will be created every 2 seconds. 21 total requests
Application protocol: h2c

finished in 12.06s, x.x req/s, y.yMB/s
requests: 42 total, 42 started, 42 done, 42 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 42 2xx, 0 3xx, 0 4xx, 0 5xx
...
```

The message `The smaller of the two will be chosen and the test will create 7 total requests.` is a bit misleading but otherwise h2load seems to send the correct expected number of requests (7 requests per client, 7*6 = 42). Or did you mean something different?

> But the code calculating these numbers are already convoluted, so might be better to clean up.

> A little bit off topic: In rate mode, it uses -m as the number of requests made per connection, which is not the original meaning of -m option.

Yes I agree the calculations are convoluted, especially when overloading the use of `-m`. Indeed, if I define a value for `-m` when using script mode and rate mode I see a different number of requests ```
h2load --threads=2 --timing-script-file='timing.script' -Bhttp://192.168.56.101:8080 --rate=2 --rate-period=1.0 --num-conn=6 -m3

-C, -n: warning: number of requests conflict. 
The smaller of the two will be chosen and the test will create 7 total requests.
starting benchmark...
spawning thread #0: 3 total client(s). Up to 2 client(s) will be created every 2 seconds. 9 total requests
spawning thread #1: 3 total client(s). Up to 2 client(s) will be created every 2 seconds. 9 total requests
Application protocol: h2c
...
```
So that was 18 total requests (6 conns * 3 reqs per connection), quite confusing!

**Comment by tatsuhiro-t** [comment]
Yes, it is scaled.  This is because rate mode adjusts number of requests to `-C` * `-m`, and `-m` defaults to the number of URIs contained in timing script. As you saw, if we change `-m`, we see different result.

I created ticket #382 to refactor rate mode configuration options.

**Comment by tatsuhiro-t** [comment]
OK, merged now.  Thank you!

**Comment by LPardue** [comment]
Great thanks!

**Comment by tatsuhiro-t** [comment]
While I was playing around --rate-period potion, I noticed that sub second rate period is possible since the value obtained from --rate-period is just used to fire timer, and there is no restriction about >= 1 second.
The easiest modification to allow < 1 in --rate-period is replace `v < 1.0` with something like `v < 1e-9`.
Alternatively, we offer rate period with time units, like `--rate-period=100ms`, as we do in nghttpx. (but in this case we cannot specify floating point number).
```
  The <DURATION> argument is an integer and an optional unit (e.g., 1s
  is 1 second and 500ms is 500 milliseconds).  Units are h, m, s or ms
  (hours, minutes, seconds and milliseconds, respectively).  If a unit
  is omitted, a second is used as unit.
```
The utility function is `util::parse_duration_with_unit`.

LPardue:
Cool!

Timing script has time periods defined as floating point `positive value of milliseconds with microsecond resolution`. At the moment rate period is `positive value representing  seconds...`. 

So the values are not consistent units (and don't necessarily need to be). However, due to the coarse granularity of the client creation I think it would be best to use period with time units as you suggest.

tatsuhiro-t:
See #384 for subsecond rate period support.

Issue #382, opened 2015-10-07
Refactor h2load rate mode options


tatsuhiro-t:
h2load rate mode works great.
The only downside is it require some thinking about how to specify the combination of -n, -C, -r, -m.
Also -m is originally the number of concurrent streams per HTTP/2 or SPDY session, that is the protocol detail configuration.  So we'd better not to use it in the way we now use it in rate mode.

I'm thinking a bit simpler configuration options, which can be used very straight forward.
But be aware that this is not backward compatible to the current rate mode.

Here is my current plan.

Without rate mode, we have no changes.
With rate mode, we only use the following options to determine the number of requests, clients, rate etc:

* -x: new option.  This is the number of requests made per connection.
* -r: This is the number of connections made per --rate-period (#380)
* -C: This is the maximum number of connections made.

We already have -r and -C.  -x is new addition.  In rate mode, -n and -c options are ignored.
To remove the test time conflicts, -C option is now mandatory if -r is used.  -x defaults to "auto", which has the same meaning of the default value of -m option.
Instead of adding -x, we can say that -n option means  the number of requests made per connection in rate mode.

What do you think about this?

tatsuhiro-t:
Perhaps, instead of adding -x, we use -n as it is now and divide by -C, which is now required option for rate mode, and calculation is trivial.

tatsuhiro-t:
More simplification can be made to remove -C option, and use -c option instead.

tatsuhiro-t:
So if we call the first plan A, then plan B is:

With rate mode, we only use the following options to determine the number of requests, clients, rate etc:

    -n: This is the total number of requests made (not per connection)
    -r: This is the number of connections made per --rate-period (#380)
    -c: This is the maximum number of connections made.

Here -c has a bit of different meaning without rate mode, but in both mode, it specifies maximum number of clients, so very similar.  If we specify -r option, h2load enters rate mode.
These 3 options are required.  They have default values, so it is not necessarily required to be in command-line.  But most likely you have to specify -c if we use -r > 1.
The number of requests made per connection is simply -n / -c.

tatsuhiro-t:
#383 implements plan B.
@nshoemaker @LPardue Do you agree this change?  Do we lose much of the split of original rate mode if we do this?

LPardue:
PR looks good. I don't think this affects how I would use h2load (other than tweaking the commands). However, I'd like to check my understanding with this change by looking at an example. 

[...]

PR #383, opened 2015-10-08
h2load: Refactor rate mode

tatsuhiro-t:
This change implements plan B described in GH-382 to simplify rate
mode, but retain its functionality as much as possible.

In this change, we removed -C option.  Instead, -c option is used to
specify the number of connections to be made, and it is now required
argument if more than 1 clients are required (this is usually the
case).  The number of requests made per connection is calculated
simply by -n / -c.

### DIFF
--- a/src/h2load.cc
+++ b/src/h2load.cc
@@ -1291,7 +1291,10 @@ void print_help(std::ostream &out) {
               scheme, host or port values.
 Options:
   -n, --requests=<N>
-              Number of requests.
+              Number of  requests across all  clients.  If it  is used
+              with --timing-script-file option,  this option specifies
+              the number of requests  each client performs rather than
+              the number of requests across all clients.
               Default: )" << config.nreqs << R"(
   -c, --clients=<N>
               Number  of concurrent  clients.   With  -r option,  this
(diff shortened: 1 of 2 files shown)


src/h2load.cc
```diff
@@ -1744,42 +1743,28 @@ int main(int argc, char **argv) {
...
+  }
 
-    if (config.nclients < config.nthreads) {
-      std::cerr << "-c, -t: the number of client must be greater than or equal "
-                   "to the number of threads." << std::endl;
-      exit(EXIT_FAILURE);
-    }
-  } else {
+  if (config.nclients < config.nthreads) {
+    std::cerr << "-c, -t: the number of client must be greater than or equal "
```
LPardue: `number of clients`

src/h2load.cc
```diff
@@ -1744,42 +1743,28 @@ int main(int argc, char **argv) {
...
+                 "to the number of threads." << std::endl;
+    exit(EXIT_FAILURE);
+  }
+
+  if (config.is_rate_mode()) {
     if (config.rate < config.nthreads) {
       std::cerr << "-r, -t: the connection rate must be greater than or equal "
```
LPardue: This restriction is still in place, with the code refactor are the calculations simplified to remove or reduce it?

For instance it would be nice to have something like `h2load -t6 -c360 -r1`, where each thread creates 60 clients at a rate of 1 client/second. This gives a ramp-up period of 60 seconds. With `-r6` the ramp up period is 10 seconds.

Its not all bad news though as a workaround at the moment is `h2load -t6 -c300 -r6 --rate-period=10`, with gives a ramp-up period of 60 seconds (as above) but with bigger steps.

src/h2load.cc
```diff
@@ -1744,42 +1743,28 @@ int main(int argc, char **argv) {
...
 
-  if (!config.is_rate_mode()) {
-    if (config.nreqs < config.nclients) {
-      std::cerr << "-n, -c: the number of requests must be greater than or "
-                << "equal to the concurrent clients." << std::endl;
-      exit(EXIT_FAILURE);
-    }
+  if (config.nreqs < config.nclients) {
+    std::cerr << "-n, -c: the number of requests must be greater than or "
+              << "equal to the concurrent clients." << std::endl;
```
LPardue: Possibly remove the word `concurrent` as this applies to both modes now?

src/h2load.cc
tatsuhiro-t: `h2load -t6 -c360 -r6 --rate-period=1` will do create 60 clients per thread at a rate of 1 client/second per thread.  --rate-period is really useful in this case.
So the discussion is whether -r option specifies rate per thread or per global.
The former will save us some calculations and remainder handling, but rest of the h2load options does not do in this way.  -t option is mostly transparent.

src/h2load.cc
tatsuhiro-t: Thank you!  Fixed now.

src/h2load.cc
tatsuhiro-t: Agreed.  Removed it now.

src/h2load.cc
LPardue: > h2load -t6 -c360 -r6 --rate-period=1 will do create 60 clients per thread at a rate of 1 client/second per thread. --rate-period is really useful in this case

Does it really do that for value `1`? It seems weird to do that it would because rate period value defaults to 1.0 if `--rate-period` is undefined, which was my first example.

> So the discussion is whether -r option specifies rate per thread or per global.
I agree this is the real point. I think it would help to update manual for `-r` to make it clear if it applies per thread or globally.

tatsuhiro-t: You are right. I completely missed the point.  Sorry for the noise.

tatsuhiro-t:
As discussed above, a5070ba implements special handling of -n option if it is used with --timing-script-file option

tatsuhiro-t:
Merged as 0d27a8991588020274eae7047b53a0f7bab0e244

PR #384, opened 2015-10-10
h2load: Support subsecond rate period

tatsuhiro-t:
This change adds subsecond rate period support to h2load.  Now
--rate-period option only accepts integer, but it can be followed by
units.  Currently, h, m, s, and ms are supported, which are hours,
minutes, seconds, and milliseconds respectively.  The underlying
functionality and usecase are already extensively used in nghttpx.

### DIFF
--- a/src/h2load.cc
+++ b/src/h2load.cc
@@ -1379,12 +1379,12 @@ Options:
               will run  as it  normally does, creating  connections at
               whatever variable rate it  wants.  The default value for
               this option is 0.
-  --rate-period=<N>
-              Specifies the time period  between creating connections.
-              The  period  must be a positive  number  greater than or
-              equal to 1.0,  representing the length of  the period in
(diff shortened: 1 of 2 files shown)


LPardue:
Tested and looks good.

I can suggest one slight improvement, where we print out the period information it would be more useful to use the same units as defined in script.

For example, On https://github.com/tatsuhiro-t/nghttp2/blob/h2load-subsecond-rate-period/src/h2load.cc#L1264 we can replace `<< std::setprecision(3) << config.rate_period << " seconds. ";` with `util::duration_str(config.rate_period) << " ";`

tatsuhiro-t:
Thank you.  That is good idea.   1245ea7 implements it.

LPardue:
Great!

tatsuhiro-t:
Thank you!  Merged via eb3505a

LPardue:
Thanks Tatsuhiro!

PR #407, opened 2015-10-25
Update doc for Debian 7.0 or above and add g++ requirement

187j3x1:
So now we can run the same command on both system, add g++ requirement maybe kind of silly, I don't know. It's your call. :D


tatsuhiro-t:
Thank you!  Merged now.

Issue #411, opened 2015-10-26
h2load: does not report the Application protocol after NPN negotiation


LPardue:
Hello, 

While testing h2load against a server that only supports NPN I noticed that the "Application protocol" line is not printed. This occurs even for a successful negotiation.

This is a minor bug that doesn't affect functionality (unless a user really wants to know the protocol).

tatsuhiro-t:
Thank you for noticing this issue.  Fix committed via b051dde

LPardue:
After testing the change I see a totally different failure:

```
starting benchmark...
spawning thread #0: 1 total client(s). 100 total requests
TLS Protocol: TLSv1.2
Cipher: ECDHE-RSA-AES128-GCM-SHA256
Server Temp Key: ECDH P-256 256 bits
No protocol negotiated. Fallback behaviour may be activated
Server does not support NPN/ALPN. Falling back to HTTP/1.1.
Application protocol: http/1.1
[ERROR] HTTP parse error: (HPE_INVALID_CONSTANT) invalid constant string

finished in 6.45ms, 0 req/s, 0B/s
requests: 100 total, 1 started, 100 done, 0 succeeded, 100 failed, 100 errored, 0 timeout
status codes: 0 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 0 bytes total, 0 bytes headers, 0 bytes data
                     min         max         mean         sd        +/- sd
time for request:        0us         0us         0us         0us     0.00%
time for connect:        0us         0us         0us         0us     0.00%
time to 1st byte:        0us         0us         0us         0us     0.00%
```

If I run this against nghttpx with log-level INFO I see

```
28/Oct/2015:10:16:08 +0000 PID5801 [INFO] shrpx_worker_process.cc:182 enc_keylen=16, hmac_keylen=32
28/Oct/2015:10:16:08 +0000 PID5801 [INFO] shrpx_worker_process.cc:241 ticket keys generation done
28/Oct/2015:10:16:08 +0000 PID5801 [INFO] shrpx_worker_process.cc:243 0 enc+dec: d3b5929ef9f77d53044cd71268236988
28/Oct/2015:10:16:08 +0000 PID5801 [INFO] shrpx_worker_process.cc:520 Entering event loop
28/Oct/2015:10:16:15 +0000 PID5801 [INFO] shrpx_connection_handler.cc:295 [LISTEN:0x7ffd582fff00] Accepted connection. fd=11
[...]

PR #415, opened 2015-10-28
h2load: do not overwrite protocol value

LPardue:
Fixes bug introduced in b051dde as part of #411


tatsuhiro-t:
Thank you.  I forget to remove that line (I added it for tests).

PR #418, opened 2015-11-06
added apparmor profile

thinred:
Hi,
I propose to add an apparmor profile for nghttpx. This is my first take, it seems to work, although it assumes that the SSL keys are stored under /etc/ssl/. I plan to add this in the Debian package, but before that I wanted to discuss this with you.

Let me know what you think!

Cheers,
Tomasz

### DIFF
--- a/contrib/usr.sbin.nghttpx
+++ b/contrib/usr.sbin.nghttpx
@@ -1,0 +1,16 @@
+#include <tunables/global>
+
+/usr/sbin/nghttpx {
+  #include <abstractions/base>
+  #include <abstractions/nameservice>
+  #include <abstractions/openssl>
+
+  capability setgid,
+  capability setuid,
+
+  /usr/sbin/nghttpx rmix,      # allow to run itself
+  /etc/nghttpx/nghttpx.conf r, # allow to read the config file
+  /etc/ssl/** r,               # give access to ssl keys
+
+  /{,var/}run/nghttpx.pid lw,  # allow to store a pid file
+}


tatsuhiro-t:
Thank you for PR!
I'm new to apparmor, so I just searched with keyword apparmor, and hit this page https://www.digitalocean.com/community/tutorials/how-to-create-an-apparmor-profile-for-nginx-on-ubuntu-14-04
It seems there are many capability settings in the example above: e.g., including abstractions/apache2-common, capability setgid, etc.
What do you think about these additional items?
It seems apache2 and nginx do not have apparmor profile in Debian.  Is there any reason they don't?

thinred:
The main idea of apparmor is to provide the bare minimum necessary to run a given program. You'll notice in my PR that a include the basic abstractions + name resolution + access to config file + SSL keys (I assume they are under /etc/ssl/) + etc. Everything that is outside this scope is going to cause nghttpx to fail (in aa-enforce mode).

The other items? I now checked that nghttpx uses setuid and setgid, so yes, we should add them (I'll update the PR in a second). The whole purpose of this PR is to find what capabilities are necessary in this profile.

Apparmor is not a part of official Debian policy (yet) and is not active by default (see https://wiki.debian.org/AppArmor/HowToUse). There are some voices tough that call for global apparmor policy.

The difficulty of providing a profile with apache2/nginx is that there are so many ways to configure these servers that having a generic profile that is also robust is hard.  The situation is much better for nghttpx which is only proxying and for the filesystem part, for example, we only need a few paths.

tatsuhiro-t:
I generally agree with you.
So apparmor prohibits anything which is not allowed in profile, it is crucial to create a profile which at least allow enough capability nghttpx performs.  It seems there is profiling tool for apparmor, I think it is good idea to run that for nghttpx.  Have you already done the profiling?

thinred:
I didn't profile using aa-logprof, but I've run nghttpx in aa-complain mode that reports all violations. It works *for me*, but arguably it may fail if you use it differently. setgid/setuid for example would fail with my first version of the profile.

tatsuhiro-t:
OK, it seems it would be hard to get it right with every possible usage scenario in the first trial, so I'll merge this PR.  Then we can start collecting and adding new stuff when we notice.

PR #433, opened 2015-11-23
h2load goes into infinite loop when timing script file starts with 0.…

shukitchan:
…0 in first line

This is a simple fix for that.
And here is my sample timing script and I run it as "h2load -t1 -r2 -c10 --timing-script-file=./script.txt"

0.0	https://screen.yahoo.com/
0.0	/__rapid-worker-1.1.js


tatsuhiro-t:
Thank you!  Merged now.

PR #435, opened 2015-11-26
fix Access to field 'sum_dep_weight' results in a dereference of a nu…

alagoutte:
…ll pointer (loaded from variable 'dep_stream') found by Clang Analyzer


tatsuhiro-t:
Thank you!
dep_stream is guaranteed to be non-null here, so I added assert(dep_stream) in aacac613af5cb0d275394830ac9cce8c457161f3, which effectively silenced scan-build.

alagoutte:
Thanks !

PR #441, opened 2015-12-04
Fix build when OpenSSL 1.0.2 is not available

sunpoet:


tatsuhiro-t:
Thank you!  Merged now.

PR #451, opened 2015-12-08
libnghttp2_asio: Added io_service accessors

apohl79:
To allow the asio wrapper to work with boost.fiber it is required
to access the underlying io_service objects.

### DIFF
--- a/src/asio_io_service_pool.cc
+++ b/src/asio_io_service_pool.cc
@@ -92,6 +92,10 @@ boost::asio::io_service &io_service_pool::get_io_service() {
   return io_service;
 }
 
+std::vector<std::shared_ptr<boost::asio::io_service>> &io_service_pool::get_io_services() {
+  return io_services_;
+}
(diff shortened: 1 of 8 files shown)


tatsuhiro-t:
Thank you!
BTW, we have http2::io_service() method to get underlying io_service object for the connection.  Do you think that is not enough to support boost fiber?   Does it need all io_service objects?

apohl79:
Yes thats true, I will update this to const. Thx!

Regarding your question. Yes - we need access to all io_service objects unfortunately, as we have to install a timer on each of them to allow a fiber scheduler on each worker thread to work.

tatsuhiro-t:
Thank you!  Merged now.

apohl79:
Great thanks!
On Dec 10, 2015 13:29, Tatsuhiro Tsujikawa <notifications@github.com> wrote:Thank you!  Merged now.

PR #456, opened 2015-12-16
libnghttp2_asio: Added access to a requests remote endpoint

apohl79:


src/asio_server_http2_handler.cc
```diff
@@ -449,6 +453,8 @@ response *http2_handler::push_promise(boost::system::error_code &ec,
 
 boost::asio::io_service &http2_handler::io_service() { return io_service_; }
 
+boost::asio::ip::tcp::endpoint http2_handler::remote_endpoint() { return remote_ep_; }
```
tatsuhiro-t: I think we can return const reference of remote_ep_ here (and all delegating functions).

tatsuhiro-t:
Thank you for PR!  Could you check my comment inline?

src/asio_server_http2_handler.cc
apohl79: You are right, fixed and pushed, thx!

tatsuhiro-t:
Thank you!  Merged now.

PR #461, opened 2015-12-25
header value should not be inp_strlower

ayanamist:
http header keys are case-insensitive, but header values are case-sensitive, so it should not be changed.


tatsuhiro-t:
Thank you for PR.  This is obviously a bug.

PR #477, opened 2016-01-12
Fix a crash in nghttpx when running on IPv6 only.

ploppy:


tatsuhiro-t:
Thank you very much!  Merged now.

PR #500, opened 2016-02-06
Don't check for dlopen/libdl on *BSD

Sp1l:
This makes linking fail on -ldl as there is no libdl on in /usr/lib or /usr/local/lib on *BSD
Tested (and part of the proposed nghttp2 1.7.0) on [https://reviews.freebsd.org/D5218|FreeBSD]


tatsuhiro-t:
Thank you very much.  Pushed and merged now.

PR #504, opened 2016-02-09
Document compiling apps and include h2load in configure

davidjb:
Finding the info on how to compile h2load was surprisingly hard -- it's got a documentation page on how to use it (https://nghttp2.org/documentation/h2load-howto.html) but compilation isn't mentioned and there's no mention in the configure file or readme.

This PR adds those details, highlighting that the ``--enable-app`` flag is what needs to be passed to `./configure` to build.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -58,9 +58,9 @@ To build the documentation, you need to install:
 
 * sphinx (http://sphinx-doc.org/)
 
-To build and run the application programs (``nghttp``, ``nghttpd`` and
-``nghttpx``) in the ``src`` directory, the following packages are
-required:
+To build and run the application programs (``nghttp``, ``nghttpd``,
+``nghttpx`` and ``h2load``) in the ``src`` directory, the following packages
+are required:
 
 * OpenSSL >= 1.0.1
 * libev >= 4.15
@@ -159,6 +159,12 @@ To compile the source code, gcc >= 4.8.3 or clang >= 3.4 is required.
    disable multi-threading in nghttpd, nghttpx and h2load to prevent
    them from crashing. A patch is welcome to make multi threading work
    on Mac OS X platform.
+
+.. note::
+
+   To compile the associated applications (nghttp, nghttpd, nghttpx and
+   h2load), you must use the ``--enable-app`` configure option and ensure
+   that the specified requirements above are met.
 
 Notes for building on Windows (Mingw/Cygwin)
 --------------------------------------------
--- a/configure.ac
+++ b/configure.ac
@@ -76,7 +76,7 @@ AC_ARG_ENABLE([threads],
 
 AC_ARG_ENABLE([app],
     [AS_HELP_STRING([--enable-app],
-                    [Build applications (nghttp, nghttpd and nghttpx) [default=check]])],
+                    [Build applications (nghttp, nghttpd, nghttpx and h2load) [default=check]])],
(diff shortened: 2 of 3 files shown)


tatsuhiro-t:
Thank you for PR!
Normally, configure script checks prerequisites and if they met, --enable-app is enabled automatically, and you don't have to manually enable it.  The only reason --enable-app exists is disable application build completely.
So if you need this option, it means probably configure script failed to detect some of the prerequisites.
The required libraries are:

*  OpenSSL >= 1.0.1
*  libev >= 4.15
*  zlib >= 1.2.3

Could you tell us which prerequisites are missing in configure summary if you invoked configure without --enable-app?  Perhaps, we can review how to detect the features.

davidjb:
The core of issue is that I wanted to compile `h2load` but how to do so wasn't clear.

https://gist.github.com/davidjb/641e3a8f557399a9fc48 contains the output of my `./configure` and `./configure --enable-app`.

That's helpful that `--enable-app` is automatically enabled if dependencies are present, but this behaviour doesn't appear to be documented anywhere; having a clear note in the README and `./configure` output solve this.  I ended up needing to go through the `Makefile`/`configure` file to determine that the `--enable-app` flag was I was looking for.   Once found and enabled via `./configure`, the error message regarding missing dependencies pointed me in the right direction.  That said, it still wasn't overly clear because the error didn't pinpoint which dependencies were missing -- as per the Gist, there are lines with `checking for [name] ... no`, but I didn't know lines are considered errors represent mandatory packages.   I ended up using a combination of this `./configure --enable-app` output and the README's dependency list to determine I was missing the following:

    libcunit1-dev libev-dev libevent-dev libjansson-dev libjemalloc-dev cython libspdylay-dev

So the general idea of my PR (which needs revising based on your additional information) was to:

* Update the README to clarify when/how to build the apps -- eg mention the auto dependency check and that ``--enable-app` flag is available.
* Update the README to mention `h2load` application gets built with the other 3 apps (see my PR).
* Change the `configure.ac` file to mention `--enable-app` will compile `h2load` (see my PR).
* The `h2load` documentation page mentions how to use the application but not how to install/compile/etc.  A link to the updated README (from first point), highlighting the automatic dependency check and the ``--enable-app`` flag, would solve that.

Separately, the mention of:
> spdylay is not packaged in Ubuntu, so you need to build it yourself: http://tatsuhiro-t.github.io/spdylay/
From Ubuntu 15.10, it is packaged in `universe` repository. See http://packages.ubuntu.com/search?keywords=libspdylay-dev

tatsuhiro-t:
OK, I agree with you.  Using --enable-app will discover and report error if dependencies are not met, which helps users a lot.  In strict sense, --enable-app is not a must (because it is enabled if all dependencies are met), but it does not work in this case.
I'll merge this PR soon.

And thank you for letting me know the spdylay package.  It was packaged in Debian, and imported to Ubuntu (recently?).  Will update README accordingly. Thanks!

davidjb:
Great, thanks.

I think just adding another short note about what you've mentioned  with
``--enable-app`` happening automatically, together with my PR, will cover
all the bases

Looks like spdylay just turned up in Wily Werewolf (15.10) and is in
testing for 16.04 -- guessing it's sticking around for a while.

tatsuhiro-t:
Done via 7adfa5dea7e0444635354b35eb47fcc4c58f2da1

PR #513, opened 2016-02-15
tests: remove unused macros

Lekensteyn:
Since v0.6.2-7-g1d138ac ("Unify DATA and other frames in
nghttp2_outbound_item and save malloc()"), the macros are unused and the
builds fails on -Werror=unused-macros.


tatsuhiro-t:
Thank you!  They were remnants since spdylay days.

PR #537, opened 2016-03-13
integration-tests: do not use recursive target

Lekensteyn:
This is a preparation for the CMake build system which does not have an "it-local" target. After this change you cannot invoke "make it" from the root directory (when using autotools), if that is wanted additional targets could be added to the top Makefile.am. (The documented `make -C integration-tests it` command is still functional though.)
___
Older automake (travis, *cough*) do not support the
AM_EXTRA_RECURSIVE_TARGETS option, this results in invoking "it-local"
directly in the travis script which relies on an implementation detail.

Since the "it" target is only used by the integration-tests directory,
just avoid the recursive targets. The README.rst suggests to enter the
integration-tests directory anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,5 +48,5 @@ script:
   - make check
   - cd integration-tests
   - export GOPATH="$PWD/integration-tests/golang"
-  - make itprep-local
-  - make it-local
+  - make itprep
+  - make it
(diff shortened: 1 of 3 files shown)


tatsuhiro-t:
Thank you.  You comment is fair enough.  Let's remove recursive `it` rules.

tatsuhiro-t:
Merged.  Thank you!

PR #538, opened 2016-03-13
Fixes and improvements for Makefile.MSVC

Jan-E:
These changes are necessary to run 'make -f Makefile.MSVC' out of the box, without manual edits. Tested on MSVC9 x86 and upwards.


Jan-E:
@tatsuhiro-t You did not add 'nghttp2_rcbuf.c' to Makefile.MSVC yet. I made use of the occasion to make a PR with some other changes in Makefile.MSVC

lib/Makefile.msvc
```diff
@@ -12,16 +12,16 @@
...
 
-_VERSION    := $(shell grep AC_INIT ../configure.ac | cut -d'[' -f3 | sed -r -e 's/(-DEV)?], //g')
+_VERSION	:= $(shell grep AC_INIT ../configure.ac | cut -d'[' -f3 | sed -e 's/-DEV//g' -e 's/], //g')
 _VERSION   	:= $(subst ., ,$(_VERSION))
 VER_MAJOR  	:= $(word 1,$(_VERSION))
 VER_MINOR  	:= $(word 2,$(_VERSION))
 VER_MICRO   := $(word 3,$(_VERSION))
 VERSION     := $(VER_MAJOR).$(VER_MINOR).$(VER_MICRO)
-VERSION_NUM := ($(VER_MAJOR) << 16) + ($(VER_MINOR) << 8) + $(VER_MICRO)
+VERSION_NUM	:= 0x$(shell if [ ${VER_MAJOR} -lt 10 ] ;then echo 0;fi)$(VER_MAJOR)$(shell if [ ${VER_MINOR} -lt 10 ] ;then echo 0;fi)$(VER_MINOR)$(shell if [ ${VER_MICRO} -lt 10 ] ; then echo 0;fi)$(VER_MICRO)
```
tatsuhiro-t: Since VER_MAJOR is decimal, when it reaches 10, this becomes 0x10, which is 16 in decimal.   So if you'd like to use 0x, you need to convert VER_MAJOR to hex string.
Of course, VER_MAJOR is not going to get bigger, but VER_MINOR could be.

tatsuhiro-t:
Thank you!  Could you check my comment?

Jan-E:
Oops. I have reverted that change, but kept the parentheses around VERSION_NUM it self. hat way it will a one line change once I have found out how to do that. Tested with latest commit on VC9 x64.

tatsuhiro-t:
Thank you.  LGTM.

PR #542, opened 2016-03-14
src/Makefile.am: fix missing allocator.h

Lekensteyn:
Fix regression from bae37e3 breaking distcheck.


tatsuhiro-t:
Thank you!  Merged now.

Issue #448, opened 2015-12-08
Request to add CMake build scripts


Type1J:
Using and debugging apps and libnghttp2 itself would be much easier if it had a CMake build system. Please consider adding a CMakeLists.txt that works for at least Windows, OSX, and Linux.

tatsuhiro-t:
I will happily accept CMake support PR.

Lekensteyn:
FYI I have started with adding CMake support. WIP at https://github.com/Lekensteyn/nghttp2/compare/master...cmake

tatsuhiro-t:
Great!

Type1J:
@Lekensteyn Thanks! This will really help me out.

wrowe:
@Lekensteyn you probably have this covered but there was a minimal implementation that might be helpful if you wanted to look;

Lekensteyn:
@wrowe Were you referring to https://github.com/trawick/nghttp2-minimal-cmake? At this moment of writing, that file is providing support to compile objects, but missing configure checks while my work has exactly the opposite problem (it implements most configure checks, but lacks sources for compilation).

I was considering to split the configure checks (~500 lines) and sources (0 lines) into separate cmake files, but maybe it's better to go ahead and finish an initial version that can actually compile stuff.

wrowe:
Yes @Lekensteyn - was just pointing this out for background.  It might be easiest time to follow Jeff's model and get the underlying lib built, then introducing the client, and server.  I'm happy to test and hopefully contribute to the effort, Trawick may be interested as well.

Lekensteyn:
Some success! I can compile and install a library and pkg-config file:

    mkdir build && cd build
    cmake .. -DENABLE_WERROR=1
    make install DESTDIR=fs

Output for `make install DESTDIR=fs`:
```
Scanning dependencies of target nghttp2
[  4%] Building C object lib/CMakeFiles/nghttp2.dir/nghttp2_map.c.o
[  9%] Building C object lib/CMakeFiles/nghttp2.dir/nghttp2_frame.c.o
[ 14%] Building C object lib/CMakeFiles/nghttp2.dir/nghttp2_queue.c.o
[ 19%] Building C object lib/CMakeFiles/nghttp2.dir/nghttp2_pq.c.o
[ 23%] Building C object lib/CMakeFiles/nghttp2.dir/nghttp2_buf.c.o
[ 28%] Building C object lib/CMakeFiles/nghttp2.dir/nghttp2_stream.c.o
[...]

PR #544, opened 2016-03-14
Add CMake build system

Lekensteyn:
This PR adds CMake support as discussed in #448. Advantages over autotools build include:

 - Faster configuration step (see below) and iterative builds (esp. with Ninja)
 - Full out-of-tree build support without cluttering source tree (with files such as configure, Makefile.in, etc.)
 - <del>In theory easier</del> support for Windows (only tested for the library with CUnit tests, not the examples or programs).
 - No static library is built nor installed, reducing the time required for `make check` (see below for times)

Notes related to CMake:

 - Only numeric versions are supported, hence the version is reported as 1.8.90 instead of 1.9.0-DEV.
 - libev 4.11 (instead of 4.15) seems to work too, the README and CMakeLists.txt file have been updated to reflect this (any reason why it said 4.15 before? Travis uses 4.11...).
 - Warnings are enabled by default (`-DENABLE_WERROR=1` will additionally add `-Werror`). Additionally, `-Wunused-macros` and `-Wstrict-prototypes` are correctly detected (autotools fails because of crap tests).
 - The `check` target is missing if cunit is not availble. Could be changed to a dummy target though.
 - <del>CMAKE_BUILD_TYPE=RelWithDebInfo is set which adds `-g -O2 -DNDEBUG` (disables assertions too!).</del>

Notable modifications to non-cmake files (`git rebase master && git diff --diff-filter=M master..`):

 - .gitignore: added cmake stuff
 - .travis.yml: added cmake+gcc/clang besides autotools+gcc/clang
 - `**/Makefile.am`: added cmake-related files to `EXTRA_DIST`
 - integration-tests/setenv.in, python/setup.py.in: added cmake-compatible libdir

distcheck is passing with this branch + #542 included (see https://travis-ci.org/Lekensteyn/nghttp2/builds/115968744), cmake can be built from the tarball.
___
Above build benchmarks were done with this script on 4d801cd, times are based on the mean of the real times, standard deviation is insignificant:
```sh
#!/bin/sh
set -u -e
git clean -xfd
autoreconf -vifs
l=logs
mkdir $l
cm='cmake -DCMAKE_C_FLAGS="-g -O2" -DCMAKE_CXX_FLAGS="-g -O2"'
for i in {1..3};do
    r='&& mkdir $_ && cd $_ && time '
    script -c "rm -rf b$r $cm         .. && time make -j10 all check" $l/make-${i}.log
    script -c "rm -rf c$r $cm -GNinja .. && time ninja     all check" $l/ninja-${i}.log
    script -c "rm -rf a$r ../configure   && time make -j10 check" $l/auto-${i}.log
done
```
times for the `check` target with cunit installed (n=3):

| what | mean | stdev
| --- | --- | ---
| ninja-config  | 7.49 | 0.020
| make-config   | 8.10 | 0.045
| auto-config   | 10.05 | 0.104
|
| ninja-build   | 27.75 | 0.274
| make-build    | 30.99 | 0.032
| auto-build    | 30.59 | 0.196

FWIW, make/ninja install takes 34ms, 500ms and 991s respectively.


Lekensteyn:
@tatsuhiro-t In 9218d5d071e67f891364fb03a7f2b2df15f515ac you added `AC_HEADER_ASSERT` "for benchmark", what benchmark? That option has as side-effect that the `--disable-assert` option becomes available.

In this PR, assertions are disabled by default (because `CMAKE_BUILD_TYPE=RelWithDebInfo` does so). To keep assertions and have debug info, use `CMAKE_BUILD_TYPE=Debug` (which is different from autotools in the sense that `-O2` is not added).

tatsuhiro-t:
Thank you for PR, I'll review this tomollow (I have a software release in another project today).

For assertions, I personally think its effect on performance in practice is negligible, so they should be enabled by default to give us useful debugging information.  FYI, autotools can use -O2 and assertion enabled at the same time.

Lekensteyn:
I also think that assertions are still useful in release builds, but the default cmake logic is apparently to disable them (that is, when you use `-DCMAKE_BUILD_TYPE=Release`). If you set `-DCMAKE_BUILD_TYPE=None -DCMAKE_C_FLAGS="-g -O2"  -DCMAKE_CXX_FLAGS="-g -O2"` you get the same result as autotools.

Should I unconditionally enable assertions for various `CMAKE_BUILD_TYPE` values or add an option for this (`-DENABLE_ASSERT=1|0`, mirrors `--enable-assert`/`--disable-assert` from autotools). I tend to prefer the former option.

tatsuhiro-t:
I also prefer the former.

tatsuhiro-t:
PR looks good.

Lekensteyn:
The assertions are fixed now (well, except on Windows where you have to define `DEBUG` explicitly for assertions).

Speaking of Windows, the library can now be built on Windows (tested with `cmake && cmake --build`). This only gives three warnings (vsnprintf usage and missing return because assertions were disabled). The main and failmalloc tests also run.

All dependencies are missing on Windows, so I had to hack a bit to get stuff working:

 - CUnit: `autoreconf -vifs && ./configure && ln -s . CUnit/Headers/CUnit` on Linux for the headers, build libcunit.lib with MSVC2013. Set `-DCUNIT_LIBRARY=\CUnit-2.1-3\VC9\Debug - Static Lib with MEMTRACE\libcunit.lib -DCUNIT_INCLUDE_DIR=\CUnit-2.1-3\CUnit\Headers`
 - zlib: extracted http://zlib.net/zlib128-dll.zip to `\zlib128` and added `-DZLIB_ROOT=\zlib128`
 - OpenSSL: took 1.0.1r from http://www.npcglib.org/~stathis/blog/precompiled-openssl/, add `-DOPENSSL_ROOT_DIR=\openssl-1.0.1r-vs2013`
 - Libev: patch for timeval (https://gist.github.com/Lekensteyn/e4586370b36d7b76d932)

Unfortunately I could not get the app to compile. The headers getopt.h and sys/un.h are missing (via utils.h) and template.h blows up completely.

Nevertheless I think that the cmake build is ready to go, others can add more Windows as desired.

tatsuhiro-t:
LGTM.  Compiling applications on Windows are known issue, currently they are *nix centric.

tatsuhiro-t:
Merged now.  Thank you for your awesome work!

PR #553, opened 2016-03-22
nghttp: Implement an Expect/Continue handshake option

jchampio:
This patchset implements an `--expect-continue` option for `nghttp`. The intent is to make it easier to test Expect/Continue handshakes against HTTP/2 server implementations. (For context, see [the definition of the Expect header](https://tools.ietf.org/html/rfc7231#section-5.1.1) as well as the last example in [RFC 7540, Sec 8.1.3](https://tools.ietf.org/html/rfc7540#section-8.1.3).)

The new option only takes effect when uploading data with `-d`. To avoid having a nonconforming server hang the client indefinitely, the Expect/Continue mechanism times out after a short delay (currently hard-coded to one second).

The three patches are divided as follows:

1. add the `getopt_long()` parsing for the new option
2. implement the handshake logic itself
3. add a timeout for each request using Expect/Continue

Feedback is very welcome, _especially_ relating to the timeout logic (and the lifetime/memory management of the new `ContinueTimer` struct), which is trickier than I wanted it to be. Thanks for your consideration!

### DIFF
--- a/src/nghttp.cc
+++ b/src/nghttp.cc
@@ -160,7 +160,8 @@ Request::Request(const std::string &uri, const http_parser_url &u,
       stream_id(-1),
       status(0),
       level(level),
-      expect_final_response(false) {
+      expect_final_response(false),
+      expect_continue(false) {
   http2::init_hdidx(res_hdidx);
   http2::init_hdidx(req_hdidx);
 }
@@ -360,10 +361,19 @@ int submit_request(HttpClient *client, const Headers &headers, Request *req) {
                                  std::string(4_k, '-'));
     }
   }
+
   auto num_initial_headers = build_headers.size();
-  if (!config.no_content_length && req->data_prd) {
-    build_headers.emplace_back("content-length", util::utos(req->data_length));
+
+  if (req->data_prd) {
+    if (!config.no_content_length) {
+      build_headers.emplace_back("content-length", util::utos(req->data_length));
+    }
+    if (config.expect_continue) {
+      req->expect_continue = true;
+      build_headers.emplace_back("expect", "100-continue");
+    }
   }
+
   for (auto &kv : headers) {
     size_t i;
     for (i = 0; i < num_initial_headers; ++i) {
@@ -401,11 +411,21 @@ int submit_request(HttpClient *client, const Headers &headers, Request *req) {
     nva.push_back(http2::make_nv_ls("trailer", trailer_names));
   }
 
-  auto stream_id =
-      nghttp2_submit_request(client->session, &req->pri_spec, nva.data(),
-                             nva.size(), req->data_prd, req);
+  int32_t stream_id;
+
+  if (req->expect_continue) {
+    stream_id = nghttp2_submit_headers(client->session, 0, -1, &req->pri_spec,
+                               nva.data(), nva.size(), req);
+  } else {
+    stream_id =
+        nghttp2_submit_request(client->session, &req->pri_spec, nva.data(),
+                               nva.size(), req->data_prd, req);
+  }
+
   if (stream_id < 0) {
-    std::cerr << "[ERROR] nghttp2_submit_request() returned error: "
+    std::cerr << "[ERROR] nghttp2_submit_"
+              << (req->expect_continue ? "headers" : "request")
+              << "() returned error: "
               << nghttp2_strerror(stream_id) << std::endl;
     return -1;
   }
@@ -1619,6 +1639,18 @@ void check_response_header(nghttp2_session *session, Request *req) {
   }
 
   if (req->status / 100 == 1) {
+    if (req->expect_continue && (req->status == 100)) {
+      int error = nghttp2_submit_data(session, NGHTTP2_FLAG_END_STREAM,
+                                      req->stream_id, req->data_prd);
+      if (error) {
+        std::cerr << "[ERROR] nghttp2_submit_data() returned error: "
+                  << nghttp2_strerror(error) << std::endl;
+        nghttp2_submit_rst_stream(session, NGHTTP2_FLAG_NONE, req->stream_id,
+                                  NGHTTP2_INTERNAL_ERROR);
+        return;
+      }
+    }
+
     req->expect_final_response = true;
     req->status = 0;
     req->res_nva.clear();
--- a/src/nghttp.cc
+++ b/src/nghttp.cc
@@ -344,7 +344,10 @@ void ContinueTimer::stop() {
 }
 
 void ContinueTimer::dispatch_continue() {
-  ev_feed_event(loop, &timer, 0);
+  // Only dispatch the timeout callback if it hasn't already been called.
+  if (ev_is_active(&timer)) {
+    ev_feed_event(loop, &timer, 0);
+  }
 }
 
 namespace {
--- a/src/nghttp.cc
+++ b/src/nghttp.cc
@@ -687,7 +687,9 @@ void HttpClient::disconnect() {
   state = ClientState::IDLE;
 
   for (auto req = std::begin(reqvec); req != std::end(reqvec); ++req) {
-    (*req)->continue_timer->stop();
+    if ((*req)->continue_timer) {
+      (*req)->continue_timer->stop();
+    }
   }
 
   ev_timer_stop(loop, &settings_timer);
--- a/src/nghttp.cc
+++ b/src/nghttp.cc
@@ -160,8 +160,7 @@ Request::Request(const std::string &uri, const http_parser_url &u,
       stream_id(-1),
       status(0),
       level(level),
-      expect_final_response(false),
-      expect_continue(false) {
+      expect_final_response(false) {
   http2::init_hdidx(res_hdidx);
   http2::init_hdidx(req_hdidx);
 }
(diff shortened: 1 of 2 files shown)


jchampio:
I think I've discovered two possible races with my timeout solution:

1. If the timeout is triggered quickly enough (i.e. before nghttp2 has sent out the headers), submission of the `DATA` frame can fail with `NGHTTP2_ERR_DATA_EXIST`.
2. If the timeout is triggered (sending the `DATA` frames) but we receive a 100 Continue interim response afterwards, the timeout callback will be triggered a second time and we'll incorrectly submit `DATA` frames again. (I had intended the `ContinueTimer` to be a single-shot, but my implementation of `dispatch_continue` invokes the timer callback even if the timer is stopped, which was unintended.)

Whether or not the first issue should be fixed is possibly up for debate, but the second issue should definitely be patched, and I'll work on a solution.

jchampio:
I've pushed a fourth patch which should fix the second race.

I'm having trouble reproducing the first race without artificially reducing the timeout to something incredibly tiny; are there good ways to force the `NGHTTP2_ERR_DATA_EXIST` race, say by delaying `HEADERS` writes?

src/nghttp.cc
```diff
@@ -304,6 +305,52 @@ void Request::record_response_end_time() {
...
+}
+
+void ContinueTimer::stop() {
+  ev_timer_stop(loop, &timer);
+}
+
+void ContinueTimer::dispatch_continue() {
+  // Only dispatch the timeout callback if it hasn't already been called.
+  if (ev_is_active(&timer)) {
+    ev_feed_event(loop, &timer, 0);
```
tatsuhiro-t: Still timer is ticking, so perhaps stop timer here?

src/nghttp.cc
```diff
@@ -414,6 +482,14 @@ int submit_request(HttpClient *client, const Headers &headers, Request *req) {
 
   req->req_nva = std::move(build_headers);
 
+  if (expect_continue) {
+    auto timer = std::make_shared<ContinueTimer>(client->loop, req);
+    timer->start();
+
+    req->continue_timer = timer;
+    client->continue_timers.push_back(timer);
```
tatsuhiro-t: I think it may be possible for Request to own std::unique_ptr of ContinueTimer (or embed timer in Request).  Request destructor will stop timer.

src/nghttp.cc
tatsuhiro-t: It would be easier to directly call callback function here instead of ev_feed_event().

tatsuhiro-t:
Thank you for the PR!  I think first race was caused because of double submission of DATA frame.  So if you fixed (2), then (1) is gone.

For expect/100-continue in general, HTTP/2 now does not need it because it has more advanced mechanism, flow control.  It has no disadvantage of waiting for the timeout where expect/100-continue has.  In HTTP/2, server can easily cancel stream with RST_STREAM, so there is not much advantage for expect dance.  Actually, expect/100-continue was not in HTTP/2 spec until the later stage.  It has been resurrected because we have no strong reason to remove it, just for compatibility for HTTP/1.1.

That said, as long as there are use cases, we can accept PR.

src/nghttp.cc
jchampio: > Still timer is ticking, so perhaps stop timer here?

The `ev_timer` is a single-shot (the repeat parameter is set to zero), so libev will automatically stop the timer as soon as it is invoked. I'll explain why I use this below:

> It would be easier to directly call callback function here instead of ev_feed_event().

My first design did this, but it had an obscure race condition: if the timer event is already pending when we get to this code, we'll end up invoking the callback twice again. You can solve that race by explicitly stopping the timer inside the callback, but at that point I was duplicating logic that libev already provided, so I moved to `ev_feed_event` since it takes care of it for us.

src/nghttp.cc
jchampio: I tried that first, but IIRC `Request` outlasts the connection itself, which means that the `ev_loop` hangs until every timeout fires.

That said... I'm still not incredibly happy with the lifetime of the `ContinueTimer`. There are two more cases where the current implementation falls short: if the server immediately sends a final response (instead of a 100 Continue), or if the stream itself is reset (say, due to a server error), the timer isn't stopped correctly.

What I'd really like is an object that is tied to the lifetime of the response on the stream, but I didn't see a good candidate. I think I will need to add explicit calls to `timer->stop()` in those two cases.

jchampio:
>  I think first race was caused because of double submission of DATA frame. So if you fixed (2), then (1) is gone.

I'm still seeing (admittedly contrived) cases where if `DATA` is submitted too soon after `HEADERS`, nghttp2 complains. I think I'm running into the behavior [documented in the Note here](https://nghttp2.org/documentation/nghttp2_submit_data.html).

> For expect/100-continue in general, HTTP/2 now does not need it because it has more advanced mechanism, flow control. It has no disadvantage of waiting for the timeout where expect/100-continue has. In HTTP/2, server can easily cancel stream with RST_STREAM, so there is not much advantage for expect dance.

Agreed. For now, the only good use case I can think of is if generation of a request body is computationally expensive or otherwise resource-intensive for a client. Something that might come up more often in the embedded space, perhaps?

Either way, my motivation for this feature is mostly to provide a tool for server implementers to check that their expect/continue handshakes still work in some sane fashion.

src/nghttp.cc
tatsuhiro-t: > so I moved to ev_feed_event since it takes care of it for us. 

Understood.

src/nghttp.cc
tatsuhiro-t: After request was submitted successfully, when stream is closed (including the case when it is reset), on_stream_close_callback() is called.  If TCP connection closed, then on_stream_close_callback may not be called, but HttpClient::disconnect() will be called.
So you can stop timer in on_stream_close_callback().  And also you can iterate requests in HttpClient::disconnect(), and stop timer for each request.  Then we don't have to use std::vector, and std::shared_ptr/std::weak_ptr.
I think if we get final response, we should stop timer, and dispatch callback.  In that point, stream may have closed, but the error will be treated by nghttp2 library gracefully.

There is one more race condition when timer is fired, before HEADERS frame is actually sent.  This may happen when the number of  concurrent streams is very low, and outstanding requests takes long time.  To workaround this, instead of starting timer just after nghttp2_submit_headers(), do it in nghttp2_on_frame_send_callback().  This ensures that HEADERS frame has been dispatched.

jchampio: > There is one more race condition when timer is fired, before HEADERS frame is actually sent. [...] To workaround this, instead of starting timer just after nghttp2_submit_headers(), do it in nghttp2_on_frame_send_callback().

Ah, I bet that's the same race I was forcing with the low timeout. I will give that (and the rest of your suggestions) a try; thanks!

jchampio:
Three patches with your most recent feedback:

1. use the `on_frame_send_callback` to start the timer
2. stop the timer outright if we get a final response or the stream is reset
3. have the `Request` own its own timer and get rid of the `shared_ptr` logic

I'm looking at one more possible race that patch 1 may have created, but I don't think nghttp itself can expose it. Then I think we'll have it. Stand by...

src/nghttp.cc
```diff
@@ -687,7 +687,9 @@ void HttpClient::disconnect() {
   state = ClientState::IDLE;
 
   for (auto req = std::begin(reqvec); req != std::end(reqvec); ++req) {
-    (*req)->continue_timer->stop();
+    if ((*req)->continue_timer) {
```
jchampio: On this note, is there a test suite for the helper utilities? I missed this typo because I ran my usual set of tests against servers, but completely forgot to run without the `--expect-continue` flag...

jchampio:
> I'm looking at one more possible race that patch 1 may have created

So far I have been unable to create a server fast enough to get this to happen, but is it possible for the `on_frame_recv_callback` to be called for a stream before that stream's `HEADERS` have been sent, say if the server responds incredibly quickly? Or does nghttp2 serialize one before the other? If this race is possible, I need to make one more change to the `ContinueTimer`.

tatsuhiro-t:
Theoretically, yes, it can happen.  nghttp2 servers don't do this though.
But I can modify nghttp client so that this never happen.  So leave it to us.

src/nghttp.cc
tatsuhiro-t: We have not had test suite for nghttp yet.  We have nghttpx integration test suite written in Go, so we can do the same sort of thing for nghttp.

jchampio:
> But I can modify nghttp client so that this never happen. So leave it to us.

Okay, great! In that case, that's the last potential issue I've been able to think of. My original three patches became eight, so let me know if you'd like me to squash or do any other history cleaning, or if you find any other issues.

src/nghttp.cc
jchampio: Go would be a new language for me, but if you'd prefer to have tests on this feature before pulling it in, I can see what I can do. No promises on how quickly it would happen, of course. :grinning:

tatsuhiro-t: Integration test for nghttp is somewhat long term goal.  No rush at the moment.  I still don't have clear picture how it will look like.

tatsuhiro-t:
Thank you!  The patch looks good.  History is short enough, so will merge as is.

tatsuhiro-t:
Now merged.  Thank you!

jchampio:
No problem; thanks for the review!

PR #581, opened 2016-04-21
in nghttp2_session_send() data is declared uninitialized and used

meconlen:
The variable data is used possibly unset after a call to 
nghttp2_session_mem_send_internal() which should
set it, however in nghttp2_session_mem_send_internal() it is
possible to return before setting the pointer.

This change initializes the variable to NULL where delcared and
sets the variable in nghttp2_session_mem_send_internal() to
NULL before possibly returning rather than after.

both options are not necessary but are both ideal practice

Note: this tickled the following compilation failure in gcc 4.8.4

nghttp2_session.c: In function 'nghttp2_session_send':
nghttp2_session.c:2942:13: error: 'data' may be used uninitialized in this function [-Werror=maybe-uninitialized]
     sentlen = session->callbacks.send_callback(session, data, (size_t)datalen,
             ^
cc1: all warnings being treated as errors

### DIFF
--- a/lib/nghttp2_session.c
+++ b/lib/nghttp2_session.c
@@ -2877,6 +2877,7 @@ static ssize_t nghttp2_session_mem_send_internal(nghttp2_session *session,
   aob = &session->aob;
   framebufs = &aob->framebufs;
 
+  *data_ptr = NULL;
   /* We may have idle streams more than we expect (e.g.,
      nghttp2_session_change_stream_priority() or
      nghttp2_session_create_idle_stream()).  Adjust them here. */
@@ -2885,7 +2886,6 @@ static ssize_t nghttp2_session_mem_send_internal(nghttp2_session *session,
     return rv;
   }
 
-  *data_ptr = NULL;
   for (;;) {
     switch (aob->state) {
     case NGHTTP2_OB_POP_ITEM: {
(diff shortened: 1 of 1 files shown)


meconlen:
Note: I found that the -O3 causes the error to be thrown on gcc 4.8.4 and gcc 5.3.0.

lib/nghttp2_session.c
```diff
@@ -3163,7 +3163,7 @@ ssize_t nghttp2_session_mem_send(nghttp2_session *session,
 }
 
 int nghttp2_session_send(nghttp2_session *session) {
-  const uint8_t *data;
+  const uint8_t *data = NULL;
```
tatsuhiro-t: Probably, this is not necessary since we always end up to call nghttp2_session_mem_send_internal, and we assign NULL to data.

tatsuhiro-t:
Thank you for the PR!  Could you check my comment in the patch?

lib/nghttp2_session.c
meconlen: I think you're correct that it's not necessary (in fact, I'm pretty sure this entire thing is a gcc bug, which I'm trying to make a test case for), but if something changes in nghttp2_session_mem_send_internal() the initialization here will keep gcc happy. 

I only included it because I I didn't look to see if anything else calls nghttp2_session_mem_send_internal() and i'm not sure if there might be any reason to change that. Either way I'm happy.

tatsuhiro-t: Thanks.  It might be explicit to set data to NULL here and nghttp2_session_mem_send instead of nghttp2_session_mem_send_internal, which would resolve your concern.
If you are OK, I'll pull this PR and do some amends and push it to the master branch.

meconlen: Sounds good.

tatsuhiro-t:
Merged.  Thank you!

PR #644, opened 2016-08-03
Make get_max_index() return the max index in frame, so we don't need

weliu:
to do extra calculation


tatsuhiro-t:
Thank you for PR.  This looks good.
One thing I noticed is that "lib/nghttp2_hd.c 100644 → 100755" is shown in "Files changed" view.  It looks like file mode is modified.  Could you check that it is intentional or not?

weliu:
Sorry I made a mistake, the file mode should keep unchanged. I will fire another pull request to fix this. Sorry about the inconvenience.

PR #647, opened 2016-08-04
Make get_max_index() return the max index in frame, so we don't need

weliu:
to do extra calculation


tatsuhiro-t:
Thank you!  Merged now.

PR #650, opened 2016-08-05
1. Modify genlibtokenlookup.py to remove redundant header comparisons.

weliu:
2. Remove inline qualifier of lookup_token() in genlibtokenlookup.py.


tatsuhiro-t:
Nice catch.  Patch looks good.  Thank you!

PR #651, opened 2016-08-09
Use memeq() instead of lstreq() in lookup_token().

weliu:
In lstreq(), we will compare the length, but this is not needed in lookup_token(). 
I have looked up the assembly code, and the gcc will optimize the code and remove this comprison during compilation. However, I still think it is the best for us to change it in code.


tatsuhiro-t:
Agreed.  I overlooked this optimization.  Nice!

PR #653, opened 2016-08-11
lib: since hd_inflate_commit_indexed() always return 0, remove the return value check in nghttp2_hd_inflate_hd_nv().

weliu:


tatsuhiro-t:
Thank you for PR.  I agree with your change.

PR #658, opened 2016-08-15
lib: Add nghttp2_hd_deflate_hd_vect() deflate API to support multiple…

weliu:
… bufs input

### DIFF
--- a/lib/includes/nghttp2/nghttp2.h
+++ b/lib/includes/nghttp2/nghttp2.h
@@ -4519,6 +4519,39 @@ nghttp2_hd_deflate_hd(nghttp2_hd_deflater *deflater, uint8_t *buf,
 /**
  * @function
  *
+ * Deflates the |nva|, which has the |nvlen| name/value pairs, into
+ * the |inlen| size of buf vector |bufsin|, each buf has length |buflen|.
+ * |buflens| is a |inlen| size of array, which will be used to store the writen
+ * size of each chunk. If one chunk is filled up, next chunk will be used.
+ * If |bufsin| is not large enough to store the deflated header block,
+ * this function fails with :enum:`NGHTTP2_ERR_INSUFF_BUFSIZE`.  The
+ * caller should use `nghttp2_hd_deflate_bound()` to know the upper
+ * bound of buffer size required to deflate given header name/value
+ * pairs.
+ *
+ * Once this function fails, subsequent call of this function always
+ * returns :enum:`NGHTTP2_ERR_HEADER_COMP`.
+ *
+ * After this function returns, it is safe to delete the |nva|.
+ *
+ * This function returns 0 if it succeeds, or one of the following
+ * negative error codes:
+ *
+ * :enum:`NGHTTP2_ERR_NOMEM`
+ *     Out of memory.
+ * :enum:`NGHTTP2_ERR_HEADER_COMP`
+ *     Deflation process has failed.
+ * :enum:`NGHTTP2_ERR_INSUFF_BUFSIZE`
+ *     The provided |buflen| size is too small to hold the output.
+ */
+NGHTTP2_EXTERN ssize_t
+nghttp2_hd_deflate_hd_vec(nghttp2_hd_deflater *deflater, uint8_t **bufsin,
+                      size_t inlen, size_t buflen, size_t *buflens,
+                      const nghttp2_nv *nva, size_t nvlen);
+
+/**
+ * @function
+ *
  * Returns an upper bound on the compressed size after deflation of
  * |nva| of length |nvlen|.
  */
--- a/lib/nghttp2_buf.c
+++ b/lib/nghttp2_buf.c
@@ -223,13 +223,62 @@ int nghttp2_bufs_wrap_init(nghttp2_bufs *bufs, uint8_t *begin, size_t len,
   return 0;
 }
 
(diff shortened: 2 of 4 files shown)


lib/nghttp2_buf.c
```diff
@@ -223,13 +223,60 @@ int nghttp2_bufs_wrap_init(nghttp2_bufs *bufs, uint8_t *begin, size_t len,
...
+  nghttp2_buf_chain *head_chain = NULL;
+
+  for (i = 0; i < in_len; ++i) {
+    uint8_t *begin = bufs_in[i];
+
+    cur_chain = nghttp2_mem_malloc(mem, sizeof(nghttp2_buf_chain));
+    if (cur_chain == NULL) {
+      while (head_chain) {
+        nghttp2_mem_free(mem, head_chain);
+        head_chain = head_chain->next;
```
tatsuhiro-t: head_chain was already freed.  Accessing head_chain->next is heap-use-after-free.

lib/nghttp2_buf.c
```diff
@@ -223,13 +223,60 @@ int nghttp2_bufs_wrap_init(nghttp2_bufs *bufs, uint8_t *begin, size_t len,
...
 void nghttp2_bufs_wrap_free(nghttp2_bufs *bufs) {
   if (bufs == NULL) {
     return;
   }
 
-  nghttp2_mem_free(bufs->mem, bufs->head);
-  bufs->head = NULL;
+  while (bufs->head) {
+    nghttp2_mem_free(bufs->mem, bufs->head);
+    bufs->head = bufs->head->next;
```
tatsuhiro-t: bufs->head was already freed.  Accessing bufs->head->next is heap-use-after-free.

lib/nghttp2_buf.h
```diff
@@ -211,6 +211,22 @@ int nghttp2_bufs_wrap_init(nghttp2_bufs *bufs, uint8_t *begin, size_t len,
                            nghttp2_mem *mem);
 
 /*
+ * Initializes |bufs| using supplied |in_len| size of buf vector |bufs_in|,
+ * each buf has length |buf_len|.
+ * The buffer size is fixed and no allocate extra chunk buffer is allocated.
```
tatsuhiro-t: Probably it should read as "no extra chunk buffer is allocated."

lib/nghttp2_hd.c
```diff
@@ -1503,6 +1503,38 @@ ssize_t nghttp2_hd_deflate_hd(nghttp2_hd_deflater *deflater, uint8_t *buf,
   return (ssize_t)buflen;
 }
 
+ssize_t nghttp2_hd_deflate_hd_vect(nghttp2_hd_deflater *deflater, uint8_t **bufsin,
+                              size_t inlen, size_t buflen, const nghttp2_nv *nv,
+                              size_t nvlen) {
```
tatsuhiro-t: We have standard struct io_vec, so perhaps, nghttp2_hd_deflate_hd_vec() sounds more like C?

lib/includes/nghttp2/nghttp2.h
```diff
@@ -4519,6 +4519,38 @@ nghttp2_hd_deflate_hd(nghttp2_hd_deflater *deflater, uint8_t *buf,
 /**
  * @function
  *
+ * Deflates the |nva|, which has the |nvlen| name/value pairs, into
+ * the |inlen| size of buf vector |bufsin|, each buf has length |buflen|.
+ *
```
tatsuhiro-t: It would be nice to tell the reader that if one chunk is used up, next chunk is used.

tatsuhiro-t:
One thing we have to work out is that nghttp2 deflate code does not necessarily use whole chunk before advancing to the next chunk.
Example: https://github.com/nghttp2/nghttp2/blob/8579b8a968226e7aa23def52effe0e968c9aedb5/lib/nghttp2_hd_huffman.c#L67

So we need some output parameter to tell the caller that how many bytes are used per chunk.

lib/nghttp2_buf.c
weliu: You are totally right, fixed it in latest PR.

lib/nghttp2_buf.c
weliu: Same as above.

lib/nghttp2_buf.h
weliu: Fixed.

lib/nghttp2_hd.c
weliu: Yes, that is more reasonable.

lib/includes/nghttp2/nghttp2.h
weliu: Fixed.

tatsuhiro-t:
Thanking about the code above, it indicates that we require at least 4 bytes per chunk.

weliu:
I have added a new function nghttp2_bufs_len_vec() in nghttp2_buf.c to output each size of buf. Thanks for your advice.

tatsuhiro-t:
Thank you.
As for the current nghttp2 implementation which might not fill up all chunk, I will see I can fix it so that it always uses all chunk provided.
Other than that, PR is good to go.

tatsuhiro-t:
Merged with some amends. Thank you!

PR #665, opened 2016-08-23
lib: Make emit_header() return void since it always succeed.

weliu:


tatsuhiro-t:
Thank for PR.  Looks like file mode is changed again.

weliu:
Sorry about that. I just correct it.

tatsuhiro-t:
Thank you!  Merged now.

PR #717, opened 2016-10-27
Fix compilation with BoringSSL

dalf:
As openssl 1.1, boringssl always enable ecdh_auto.

https://boringssl.googlesource.com/boringssl/+/2357%5E!/


tatsuhiro-t:
Thank you for the PR.  Merged.

PR #727, opened 2016-11-11
Prevent undefined behavior in decode_length

ruds:
When doing some fuzz testing, I found inputs that caused decode_length to be called in such a way that shift grew to values larger than 32. This is undefined behavior.


tatsuhiro-t:
Thank you for PR.  I confirmed that nghttp2 may decode wrong number when this is exploited.

PR #740, opened 2016-12-10
nghttpx: Dynamic DNS

### LINKED ISSUE #721: Is nghttpx bind to IP or address?
I'm using nghttpx as a client on my macbook and connecting to server using H2 tunnel.

It seems that nghttpx still bind to server old IP even when my domain dns record changed
```
backend=mydomain.com,443;;proto=h2;tls
```

Is this expected behavior? How can I make nghttpx bind to new address when domain record changed?

tatsuhiro-t:
This PR adds dynamic DNS to nghttpx.
Previously, name lookup for backend servers are done at start up, or configuration reloading, and after that, nghttpx keeps using them.
With this PR, nghttpx can resolve host name dynamically.  Initial name lookup is skipped.
We use c-ares library to perform asynchronous name lookup.  As of this PR, c-ares is added to the required packages to build nghttpx.

To use this feature, use "dns" parameter in backend option.  The feature is disabled by default, and needs to be enabled explicitly:

```
backend=example.com;;dns
```

Added couple of new options: --dns-cache-timeout, --dns-lookup-timeout, and --dns-max-try

Closes #721


tatsuhiro-t:
No, it only uses first one.

PR #769, opened 2017-01-03
Fix issue reporting by PVS Studio

alagoutte:
I have make a pass of PVS Studio (www.viva64.com) on nghttp2 and fix some issue reported (not found major issue...)

http://www.viva64.com/en/m/0036/


tatsuhiro-t:
Thank you!  This is neat and nice.

PR #783, opened 2017-01-16
Fix authority for --get-assets if IP address is used 

worenga:
Fix authority for --get-assets if IP address is used in conjunction with user-defined :authority header.

Note that the behavior of --get-assets/-a differs when used with a IP, if the website does not link to the IP but the authority.

Example: 
-------------
Assume the page links an image with the absolute url ``

`$nghttp https://www.example.com -nsa --no-push`
will download image.jpg
`nghttp https://xxx.xxx.xxx.xxx -H':authority:www.example.com' -nsa`
where `xxx.xxx.xxx.xxx` is the resolved IPv4 address for `www.example.com`, will not download image.jpg

### DIFF
--- a/src/nghttp.cc
+++ b/src/nghttp.cc
@@ -1596,20 +1596,32 @@ void update_html_parser(HttpClient *client, Request *req, const uint8_t *data,
     auto res_type = p.second;
 
     http_parser_url u{};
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Thank you for PR.  Cherry picked as 25503ad7 with some amends.

PR #791, opened 2017-02-01
add support for link rel="preload" for --get-assets

worenga:
This pull requests adds support for link rel="preload" resources in the html [1].
Previously, nghttp used with --get-assets did not download resources indicated to preload.
This behavior is closer to the behavior of a real-world browser such as Firefox.

[1] https://www.w3.org/TR/preload/


tatsuhiro-t:
Thank you for PR!  I will review it.

tatsuhiro-t:
Merged now.

PR #794, opened 2017-02-03
Acceptor infinite loop

clemahieu:
start_accept is called even if the acceptor is closed causing an infinite loop.


tatsuhiro-t:
Thank you for PR.  I'd like to cherry pick a722dc12ac80f2f2d9ba65e61d2258b673ca9625 only.
 bcc29cd needs separate discussion, and it also requires to show a way to achieve multi threaded server easily.

tatsuhiro-t:
Cherry picked as 6d3e010ae7f49b14cb1201367e5479a86d9d8e94.

PR #795, opened 2017-02-03
close_stream erases from streams_ while it's being iterated over.

clemahieu:
close_stream just erases the id from the streams_ container which invalidates the iterator.  The who streams_ data structure is about to be deleted anyway so this is unnecessary.


tatsuhiro-t:
Thank you for PR.  Confirmed that this is indeed a real bug .

tatsuhiro-t:
Thank you.  Merged now.

PR #797, opened 2017-02-03
Using io_service passed in to server rather than managing our own.

clemahieu:
This is a separate pull request which modifies the nghttp2 asio server api to be more in line with boost asio conventions.

The boost::io_service class is designed to allow multiple threads to service io operations.  While it's possible to use multiple io_services to process requests, this is often unnecessary.  If multiple io_services were necessary I believe their lifetimes should be managed by the enclosing program itself, rather than by the server library itself which may only be one of many libraries using asio in the same process.  nghttp2 as-is requires new io_services precluding the ability for the enclosing program to use already existing io_services.

This documentation has a related quote.  https://theboostcpplibraries.com/boost.asio-scalability "Calling run() repeatedly on a single I/O service object is the recommended method to make a program based on Boost.Asio more scalable. However, instead of providing several threads to one I/O service object, you could also create multiple I/O service objects."

The biggest issue I see with the implementation as-is is with the server creating its own threads.  Program writers looking to integrate nghttp2 very well may already have an io_service and a thread pool for other asio operations and wish to simply add nghttp2 operations to this service.  Having nghttp2 manage the thread pool also precludes the ability to do fine-grained operating system specific tuning of thread parameters for better io performance instead of default thread parameters.


tatsuhiro-t:
Thank you for PR.
I read the linked document, and understand how to use threads with one io_service.
One important thing that libnghttp2_asio requires is that all operations/callbacks for one particular connection must be done in a same thread at a time.  I did some experiments, but it seems that running sincle io_service in several threads does not satisfy the above requirement.
https://github.com/nghttp2/nghttp2/tree/asio_threading is my branch for the experiment.  I added some commits to compile with autotools and examples/asio-sv and examples/asio-sv2 now compile.

clemahieu:
I'll take a look and see what I can do.

Thanks again for making this library, this is much better than any other c++ http library I've used.

andrzejc:
@tatsuhiro-t Hello, what is the status of this PR, is it ever planned to be integrated? I was planning to contribute the same feature to libnghttp2_asio when I noticed this PR. I looked through the code here and I see that it may be problematic in a way that it replaces the existing thread pool, making the change incompatible with existing consumers of nghttp2, is this correct? My approach would be to extend the current API offering the use of user-provided io_context as an alternative, without breaking the existing code. I could contribute the changes to make it work this way, however I'd like to know what are the chances they would be accepted into mainline nghttp2 library. I prefer to have this feature integrated rather than having to maintain my own fork and have to keep the for in sync with upstream development.

tatsuhiro-t:
I have serious maintenance issues for nghttp2 asio library.  If you are interested in improving or hacking the library, it would be better to fork it yourself.

andrzejc:
> I have serious maintenance issues for nghttp2 asio library. If you are interested in improving or hacking the library, it would be better to fork it yourself.

ok, thanks for the reply

tatsuhiro-t:

PR #803, opened 2017-02-09
Holding more shared_ptrs instead of raw ptrs

clemahieu:
This makes sure objects don't get deleted that are referenced by callbacks inside io_service.

This seems to fix memory issues from my tests with valgrind and -fsanitize=address.  Do you see other errors?  Do you have other tests I should run?


tatsuhiro-t:
Thank you for PR.  I think your fix is valid.  I always use address sanitizer these days to test memory related issue.

tatsuhiro-t:
I meat to merge this into master.

tatsuhiro-t:
Cherry picked into master as 298808f27613c2e81175f64b753a190bd8b598ec

PR #805, opened 2017-02-10
graceful stop of nghttp2::asio_http2::server::http2

pakdel:
Explicit io_service::stop() will prevent running streams from finishing their task. That means if there are  reponses that we have already called end(std::string) on them and they have not finished sending back their data, they will be closed with a NGHTTP2_INTERNAL_ERROR

Instead, we can stop accepting connections and destroy all io_service::work objects to signals end of work.

### DIFF
--- a/src/asio_io_service_pool.cc
+++ b/src/asio_io_service_pool.cc
@@ -75,11 +75,16 @@ void io_service_pool::join() {
   }
 }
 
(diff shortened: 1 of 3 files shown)


tatsuhiro-t:
Thank you for PR!  Merged now.

PR #807, opened 2017-02-10
nghttpx: Send SIGQUIT to the original master process

### LINKED ISSUE #804: USR2 should gracefully shutdown old processes
I think sending USR2 should also turn old processes into graceful shutdown mode.
Right now, USR2 will cause new processes to be spawned. Old ones sticks around until administrator clean them up with QUIT.
Doing `ipc_send(wp, SHRPX_IPC_GRACEFUL_SHUTDOWN);` in old master after spawning new processes would make the old processes auto clean up after all connections are drained.

Was there a reason to leave old processes after connections have been serviced and new processes were spawned?

tatsuhiro-t:
Previously, after sending SIGUSR2 to the original master process, and
the new master process gets ready, user has to send SIGQUIT to the
original master process to shut it down gracefully.  With this commit,
the new master process sends SIGQUIT to the original master process
when it is ready to serve requests, eliminating for user to send
SIGQUIT manually.

This works nicely with systemd, because now you can replace nghttpx
binary with new one by "systemctl kill -s USR2 nghttpx".

Fixes #804

### DIFF
--- a/src/shrpx.cc
+++ b/src/shrpx.cc
@@ -123,6 +123,12 @@ constexpr auto ENV_UNIX_PATH = StringRef::from_lit("NGHTTP2_UNIX_PATH");
 // descriptor.  <PATH> is a path to UNIX domain socket.
 constexpr auto ENV_ACCEPT_PREFIX = StringRef::from_lit("NGHTTPX_ACCEPT_");
 
+// This environment variable contains PID of the original master
+// process, assuming that it created this master process as a result
+// of SIGUSR2.  The new master process is expected to send QUIT signal
+// to the original master process to shut it down gracefully.
+constexpr auto ENV_ORIG_PID = StringRef::from_lit("NGHTTPX_ORIG_PID");
+
 #ifndef _KERNEL_FASTOPEN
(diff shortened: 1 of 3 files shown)


tatsuhiro-t:
Well, we need --kill-who=main to send SIGUSR2 to only "main" nghttpx process (e.g., `systemctl kill -s USR2 --kill-who=main nghttpx`).  Otherwise each exiting master process spawn its own new process.

src/shrpx.cc
```diff
@@ -1311,6 +1338,12 @@ int event_loop() {
   // ready to serve requests
   shrpx_sd_notifyf(0, "READY=1");
 
+  if (orig_pid != -1) {
+    LOG(NOTICE) << "Send QUIT signal to the original master process to tell "
+                   "that we are ready to server requests.";
```
zdzichu: 'serve', not 'server'

tatsuhiro-t: Thank you.  Fixed.

tatsuhiro-t:
Squashed and fixed commit comment.  Ready to merge.

PR #814, opened 2017-02-14
nghttp2_session: fix The 'then' statement is equivalent to the subseq…

alagoutte:
…uent code fragment found by PVS Studio (V523)


tatsuhiro-t:
Thank you for PR. Looks good to me.

PR #823, opened 2017-02-27
nghtt2_gzip: fix this statement may fall through [-Werror=implicit-fa…

alagoutte:
…llthrough=] found by gcc7


tatsuhiro-t:
Thank you for PR.  Looks good to me.

PR #825, opened 2017-02-28
Added --with-openssl=PATH option to configure to allow for user-defin…

mbrand1:
…ed location of OpenSSL

I wanted a way to specify a different OpenSSL location.  It prepends the location of OpenSSL's openssl.pc file to PKG_CONFIG_PATH so pkg-config takes flags from there instead of the distro's version.  I basically lifted this from the Apache source.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -142,6 +142,11 @@ AC_ARG_WITH([cython],
                     [Use cython in given PATH])],
     [cython_path=$withval], [])
 
+AC_ARG_WITH([openssl],
+    [AS_HELP_STRING([--with-openssl=PATH],
+                    [Use openssl in given PATH])],
+    [openssl_path=$withval], [])
+
 dnl Define variables
 AC_ARG_VAR([CYTHON], [the Cython executable])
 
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Thank you for PR.

What you did in PR is already possible with the following command line:

```
$ ./configure PKG_CONFIG_PATH=/path/to/openssl/lib/pkgconfig
```

This is basic facility provided by pkg-config macro.

mbrand1:
Yes, I see that you could just set the var on the command line, but I'm not sure most would be aware of that.  Is it not desirable to have an option (or at least document that specifically)?

What drove me to this is I primarily use Debian which still distributes OpenSSL 1.0.1x.

So the only way to make this work is to point to a custom-installed OpenSSL.  I didn't see this documented, and thought it made sense to have an option for it.

tatsuhiro-t:
./configure -h shows this:
```
  PKG_CONFIG_PATH
              directories to add to pkg-config's search path
```

Other than that, we have no documentation for it.  The reason why we don't have is I thought that it is rather standard practice with pkg-config usage.  If it is not, I'm happy to add some text to README to guide user in this path.

I regularly uses different OpenSSL implementations (e.g., BoringSSL, LibreSSL, etc) which are installed in custom locations to test nghttp2 applications (these OpenSSL forks have slightly different API, and  bothers me a lot..).  I always use `./configure PKG_CONFIG_PATH=...` to build nghttp2 with one of them.

mbrand1:
Yes I think this would be helpful in the README to explain how you are compiling with different implementations of SSL.

tatsuhiro-t:
I added some text to explain how to use PKG_CONFIG_PATH with configure script in dc7c12d4745ca7a207c0fe061f58c61d740599a1

mbrand1:
Ok I can accept using PKG_CONFIG_PATH is the way to go for this..

PR #881, opened 2017-04-11
Support specifying stream priority via session::submit()

mway:
i noticed that the C++ API doesn't pass along priority (just passes `nullptr` currently), so i thought a clean way to do this (without exposing the underlying nghttp2 types) would be to just create a simple wrapper that could be passed along, whose sentinel value would result in the same `nullptr` argument to `nghttp2_submit_request` (i.e. no specified priority), but when properly constructed, would pass an `nghttp2_priority_spec*`.

for convenience, it also seemed reasonable that `priority_spec` ctor should be easily usable by providing the most common arguments first, i.e., trivially used to represent weight (e.g. `priority_spec(4.0)`).

i need this on my end, so happy to make any adjustments you'd like!

### DIFF
--- a/src/asio_client_session.cc
+++ b/src/asio_client_session.cc
@@ -96,28 +96,47 @@ boost::asio::io_service &session::io_service() const {
 
 const request *session::submit(boost::system::error_code &ec,
                                const std::string &method,
-                               const std::string &uri, header_map h) const {
-  return impl_->submit(ec, method, uri, generator_cb(), std::move(h));
+                               const std::string &uri, header_map h,
+                               priority_spec prio) const {
+  return impl_->submit(ec, method, uri, generator_cb(), std::move(h),
+                       std::move(prio));
 }
 
 const request *session::submit(boost::system::error_code &ec,
                                const std::string &method,
                                const std::string &uri, std::string data,
-                               header_map h) const {
+                               header_map h, priority_spec prio) const {
   return impl_->submit(ec, method, uri, string_generator(std::move(data)),
-                       std::move(h));
+                       std::move(h), std::move(prio));
 }
 
 const request *session::submit(boost::system::error_code &ec,
                                const std::string &method,
                                const std::string &uri, generator_cb cb,
-                               header_map h) const {
-  return impl_->submit(ec, method, uri, std::move(cb), std::move(h));
+                               header_map h, priority_spec prio) const {
+  return impl_->submit(ec, method, uri, std::move(cb), std::move(h),
+                       std::move(prio));
 }
 
 void session::read_timeout(const boost::posix_time::time_duration &t) {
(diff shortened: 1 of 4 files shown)


src/asio_client_session.cc
```diff
@@ -96,29 +96,51 @@ boost::asio::io_service &session::io_service() const {
...
+  spec_.exclusive = exclusive;
+}
+
+const nghttp2_priority_spec *priority_spec::get() {
+  if (!valid_) {
+    return nullptr;
+  }
+
+  nghttp2_priority_spec_init(&spec_, spec_.stream_id, spec_.weight,
+                             spec_.exclusive);
```
tatsuhiro-t: I think it would be a bit nicer to do this in ctor, rather than here.

src/asio_client_session.cc
```diff
@@ -96,29 +96,51 @@ boost::asio::io_service &session::io_service() const {
...
+                               header_map h, priority_spec prio) const {
+  return impl_->submit(ec, method, uri, std::move(cb), std::move(h),
+                       std::move(prio));
 }
 
 void session::read_timeout(const boost::posix_time::time_duration &t) {
   impl_->read_timeout(t);
 }
 
+priority_spec::priority_spec(int32_t weight, int32_t stream_id, int exclusive)
```
tatsuhiro-t: I think it would be nice to arrange the order of the parameter in the same order as nghttp2_priroty_spec_init, that is, stream_id, weight, and then exclusive.
Since exclusive is effectively boolean, I prefer bool type for exclusive to generic int.

mway: yeah - i'd gone back and forth with consistency because it seemed like, for the simplest use, the weight would be the only specified parameter. but happy to keep it the same.

src/asio_client_session.cc
mway: sounds good to me!

src/asio_client_session.cc
```diff
@@ -96,29 +96,53 @@ boost::asio::io_service &session::io_service() const {
...
 void session::read_timeout(const boost::posix_time::time_duration &t) {
   impl_->read_timeout(t);
 }
 
+priority_spec::priority_spec(const int32_t stream_id, const int32_t weight,
+                             const bool exclusive)
+    : valid_(true) {
+  spec_.weight = weight;
+  spec_.stream_id = stream_id;
+  spec_.exclusive = exclusive;
```
tatsuhiro-t: I think these 3 lines can be removed.  nghttp2_priority_spec_init basically does that.

mway: whoops! meant to remove that. thanks for catching.

tatsuhiro-t:
Thank you!  Merged now.

PR #897, opened 2017-04-18
Maintain full HTTP/1.1 support in forward proxy mode

bemasc:
Currently, HTTP/1.1 requests are only supported in
forward proxy mode if they contain an absolute URI path.
With this change, HTTP/1.1 requests with a local path
are treated like an equivalent HTTP/2 request, i.e. relying
on the Host header to indicate the destination domain.

### DIFF
--- a/src/shrpx_https_upstream.cc
+++ b/src/shrpx_https_upstream.cc
@@ -383,6 +383,7 @@ int htp_hdrs_completecb(http_parser *htp) {
 
       if (host) {
         req.authority = host->value;
+        req.no_authority = false;
       }
 
       if (handler->get_ssl()) {


tatsuhiro-t:
Thank you for PR.
According to https://tools.ietf.org/html/rfc7230#section-5.3.2, client MUST use absolute form to talk to a proxy.  The sane browsers already do this.  Could you share us the motivation to enable non-standard practice?

bemasc:
Many forward proxy backends, such as Apache HTTPD, support both relative and absolute queries.  This allows the proxy to also act as a web server, on the same port.  This can be convenient, e.g. to provide a page of instructions on how to configure the client's proxy settings.

This PR lets me use nghttpx in front of Apache in that configuration.

So I don't think this is "non-standard", rather it is for being both a "standard proxy" and a "standard web server".

tatsuhiro-t:
Thank you.  I understand your use case.
Since nghttpx runs in either reverse proxy or forward proxy, I'm not sure it is good idea to allow non-absolute URI for forward proxy just because it is convenient or just works.
If it is instructed to operate as forward proxy, it should do some validations, and only absolute URI should be allowed.
For example, squid does not allow non-absolute URI.

bemasc:
Apache HTTPD and Apache Traffic Server both support this "hybrid mode", but unfortunately it's not currently possible to use it through nghttp2.  That's what I'm trying to solve.  If I could turn "forward" and "reverse" mode on at the same time, that would also solve it, but currently nghttp2 doesn't allow that.

I don't think there's a spec compliance issue at all.  RFC 7230 only applies to HTTP clients, not HTTP servers.  The server here is a fully spec compliant HTTP server. There's also no security issue; a malicious client doesn't gain any advantage.

PR #901, opened 2017-04-24
ORIGIN frame support

LPardue:
This is a very rough 1st attempt at adding ORIGIN frame support natively into nghttp2, based on https://tools.ietf.org/html/draft-ietf-httpbis-origin-frame-03. 

More work is required but I wanted to gauge your opinion on adding this natively to the library before I flesh out tests etc.


tatsuhiro-t:
Thank you.  I'm interested in native ORIGIN frame support, and would like to see it in libnghttp2.
I'm not following this draft, not sure this is going to be ready to publication of RFC.  But I'd like to wait for RFC before merging this PR (I know this is still WIP) so that the libnghttp2 API is kept stable.

LPardue:
The frame format is quite simple, like a simpler ALTSVC. The complexity is in logic required for applications, which is still subject to discussion (see IETF 98 minutes at - https://www.ietf.org/proceedings/98/minutes/minutes-98-httpbis-00).

I think holding this from mainline nghttp2 until it emerges as an RFC is the right thing to do. I'm hoping to flesh it out fully,  so that implementations could start experimenting with ORIGIN quite easily if they wanted.

tatsuhiro-t:
Experimentation is good thing.  You can also add a functionality to nghttpd server to send ORIGIN frame so that client developer can see what's happening.

In ALTSVC, I decided to just pass whole payload to application.  This is because the content of the frame is just the value of alt-svc header field.  ORIGIN frame has several repeatable length prefixed fields, so it might be nicer to parse them.  Is the frame format still moving target?

LPardue:
I think they simplified the frame from previous drafts. Not sure how settled it is now. The frame format for latest 03 draft is 

~~~
The ORIGIN frame type is 0xc (decimal 12).

   +-------------------------------+-------------------------------+
   |         Origin-Len (16)       | ASCII-Origin? (*)           ...
   +-------------------------------+-------------------------------+

   The ORIGIN frame's payload contains the following fields, sets of
   which may be repeated within the frame to indicate multiple origins:

   Origin-Len:  An unsigned, 16-bit integer indicating the length, in
      octets, of the ASCII-Origin field.

   Origin:  An optional sequence of characters containing the ASCII
      serialization of an origin ([RFC6454], Section 6.2) that the
      sender believes this connection is or could be authoritative for.

   The ORIGIN frame does not define any flags.  However, future updates
   to this specification MAY define flags.  See Section 2.2.
~~~

LPardue:
So the test `session_recv_origin` fails 
~~~
Test: session_recv_origin ...FAILED
    1. nghttp2_session_test.c:2294  - 1 == ud.frame_recv_cb_called
    2. nghttp2_session_test.c:2295  - NGHTTP2_ORIGIN == ud.recv_frame_hd.type
~~~

This is due to the following size check in `session_mem_recv`  failing.

```
if (2 + origin_len > iframe->payloadleft)
```

 I must be doing something wrong but can't quite pinpoint it.

tatsuhiro-t:
Assuming that that statement is inspired from altsvc code, "2 +" is not required for ORIGIN frame.

In fact, it is a bit vague at this moment why I chose "2" in altsvc.  It probably because Alt-Svc-Field-Value cannot be empty, and it requires more than 2 characters in its definition.  But then 4 is more suitable value than 2.

BTW, the draft indicates that  (Origin-Len, ASCII-Origin?) pair can be repeated:

>    The ORIGIN frame's payload contains the following fields, sets of
   which may be repeated within the frame to indicate multiple origins:

It seems your PR assumes the frame contains just one pair.

LPardue:
Oh yes you are quite right. I'll try to revise the code to accommodate this.

LPardue:
And I'll also think about a structure to present the parsed data to library caller.

LPardue:
Still have this in mind but it's fallen behind some other tasks I have to do.

tatsuhiro-t:
There is no rush, of course.  Origin frame is not finalized yet, so there is much time to left.

LPardue:
Something horrible has happened to this PR in trying to bring it up to date. Given that my first attempt was wrong anyway, I'm inclined to file a new PR with a better implementation.

jasnell:
Just wanted to check in on this. @LPardue, are you still planning to work on this?

LPardue:
Planning to yes, finding time is trickier - always helps to have a use
case. Do you have something in mind? In the intervening time I've become
more familiar with node...

On 29 Dec 2017 19:51, "James M Snell" <notifications@github.com> wrote:

Just wanted to check in on this. @LPardue <https://github.com/lpardue>, are
you still planning to work on this?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/nghttp2/nghttp2/pull/901#issuecomment-354492764>, or mute
the thread

jasnell:
Nothing specific. Just planning ahead for node.js' impl

jasnell:
@LPardue ... just following up on this. The key use case we have for this in Node.js is in the implementation of a client-side connection/session pool. We want the server to be able to declare it's origin set so that the selection code on the client side can know which origins an existing connection is authoritative for.

Aside from the rebase issues here, the one major issue with the current implementation in this PR is that it assumes one origin per ORIGIN frame, but the current version of the spec (the one waiting to be published as an RFC) allows zero-to-many origins per ORIGIN frame, which makes unpacking a bit more difficult because there's no way to quickly calculate how many discreet origins there are in any given ORIGIN frame that you receive.

@tatsuhiro-t ... I wanted to ask your advice on that particular piece. Specifically, the payload of any given ORIGIN frame is N origin-entry records, where an origin-entry record is a 16-bit length L followed by L bytes. How would you prefer to handle the memory allocations for parsing this?

tatsuhiro-t:
@jasnell I am fine to allocate memory to store multiple origins per frame as we did for SETTINGS frame.  Assuming that ORIGIN frame is not frequently received, its overhead is minimal.

jasnell:
The one difference from the SETTINGS frame is that origins are variable length and we have no way of knowing up front how many origin entries to allocate for based on the payload length.

tatsuhiro-t:
Yes, but we first parse payload to just know the number of origins, and
then allocate array, and fill them in 2nd pass.

2018/01/02 16:37 "James M Snell" <notifications@github.com>:

jasnell:
Yep, works for me. That's the direction I was going but wanted to make sure you were good with it.

LPardue:
Regarding the single origin issue, yes that was due to a mistake I made in
reading the specification. The author kindly improved the clarity of the
spec in response to my feedback.

I may start a new PR to approach ORIGIN differently and to sidestep bit rot
on this one.

LPardue:
It's unlikely I will find the time to get around to this any time soon. Happy for someone else to take the reigns.

tatsuhiro-t:
I'll take over the implementation of Origin frame.

tatsuhiro-t:
Closed in favor of #1177

PR #921, opened 2017-05-18
travis: Use nproc to speed up build (potentially)

tatsuhiro-t:


tatsuhiro-t:
gcc stopped with internal error.  I'm not sure the parallel make caused it.

PR #938, opened 2017-06-11
fix cleaning in out-of-tree builds

benjaminp:
The altered line previously failed if the rst sources hadn't been copied over.


tatsuhiro-t:
Thank you!  Merged now.

PR #965, opened 2017-07-21
Adding timing-sensitive load test option in h2load.

sohamm17:
Arguments are not well checked right now. Need suggestions. So to run our timing test, we need to run:
`./src/h2load -n 0 -c 5 -r 5 --rate-period 10 -t 1 --warm-up-time 5 https://localhost:4500`
This will create 5 clients and have 5 seconds of warm-up time. The main test duration is given as the `rate-period` argument, in this case, which is 10 seconds. Number of connections (`n`) should be `0`. This helps us to know that infinite number of connections are to be tried. Currently, multiple threads are not tried. So, `t` should be `1`.
We also need to check whether previous benchmarks are working fine.

### DIFF
--- a/src/h2load.cc
+++ b/src/h2load.cc
@@ -90,6 +90,7 @@ Config::Config()
       connection_window_bits(30),
       rate(0),
       rate_period(1.0),
+      warm_up_time(0.0),
       conn_active_timeout(0.),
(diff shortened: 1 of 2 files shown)


sohamm17:
Sorry. Did this mistakenly. Please ignore.

PR #968, opened 2017-07-28
Enabling timing-based load-testing in h2load

sohamm17:
This option will enable timing-based load-testing in h2load. Users can specify warm-up and main duration of the load-test. There will be no finite number of requests given.` h2load` will try to generate as many requests as possible, both during warm-up and the main duration period. However, statistics such as the total `req/s` will be calculated only for the main duration period.
For this test to run, we can invoke h2load by the following command, for example:
`h2load -c 10 -t 2 -D 5 --warm-up-time 5 -n 0 https://localhost:4500`

The above says that there will be 10 clients for 2 threads with 5 seconds of warm-up time and 5 seconds of main duration of load-test. `-n` needs to be set to zero for timing-based test.

### DIFF
--- a/src/h2load.cc
+++ b/src/h2load.cc
@@ -193,6 +193,7 @@ void writecb(struct ev_loop *loop, ev_io *w, int revents) {
     rv = client->connect();
     if (rv != 0) {
       client->fail();
+      client->worker->free_client(client);
       delete client;
       return;
     }
@@ -200,6 +201,7 @@ void writecb(struct ev_loop *loop, ev_io *w, int revents) {
   }
   if (rv != 0) {
     client->fail();
+    client->worker->free_client(client);
     delete client;
   }
 }
@@ -213,6 +215,7 @@ void readcb(struct ev_loop *loop, ev_io *w, int revents) {
     if (client->try_again_or_fail() == 0) {
       return;
     }
+    client->worker->free_client(client);
     delete client;
     return;
   }
@@ -245,9 +248,10 @@ void rate_period_timeout_w_cb(struct ev_loop *loop, ev_timer *w, int revents) {
       client->fail();
     } else {
       if (worker->config->is_timing_based_mode()) {
-        worker->clients.push_back(client.get());
+        worker->clients.push_back(client.release());
+      } else {
+        client.release();
       }
-      client.release();
     }
     worker->report_rate_progress();
   }
@@ -268,10 +272,12 @@ void duration_timeout_cb(struct ev_loop *loop, ev_timer *w, int revents) {
   auto worker = static_cast<Worker *>(w->data);
 
   for (auto client: worker->clients) {
-    client->req_todo = client->req_done; // there was no finite "req_todo"
-    worker->stats.req_todo += client->req_todo;
-    client->req_inflight = 0;
-    client->req_left = 0;
+    if (client) {
+      client->req_todo = client->req_done; // there was no finite "req_todo"
+      worker->stats.req_todo += client->req_todo;
+      client->req_inflight = 0;
+      client->req_left = 0;
+    }
   }
 
   worker->current_phase = Phase::DURATION_OVER;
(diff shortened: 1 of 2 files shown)


sohamm17:
@htuch

tatsuhiro-t:
Thank you.  I'm a bit busy these days for the other project.  I will review this PR when I have time.

src/h2load.cc
```diff
@@ -1176,37 +1298,62 @@ Worker::Worker(uint32_t id, SSL_CTX *ssl_ctx, size_t req_todo, size_t nclients,
...
+    stats.client_stats.reserve(std::min(nclients, max_samples));
+  }
 
   sampling_init(request_times_smp, req_todo, max_samples);
   sampling_init(client_smp, nclients, max_samples);
+
+  if (config->is_timing_based_mode()) {
+    duration_watcher.data = this;
+
+    ev_timer_init(&warmup_watcher, warmup_timeout_cb, config->warm_up_time, 0.);
```
tatsuhiro-t: It might be better to call ev_timer_init even if config->is_timing_based_mode() returns false so that we always know that it is initialized, and we can stop it in dtor.   worker->duration_watcher is also initialized here.
Just start it if we are in timing based mode.

tatsuhiro-t:
When I stop server while h2load is warming up, it crashes:

```
$ src/h2load -c10 -D10 --warm-up-time=5 -n0 http://localhost:3000
starting benchmark...
spawning thread #0: 10 total client(s). Timing-based test with 5s of warm-up time and 10s of main duration for measurements.
Warm-up started for thread #0.
progress: 10% of clients started
progress: 20% of clients started
progress: 30% of clients started
progress: 40% of clients started
progress: 50% of clients started
progress: 60% of clients started
progress: 70% of clients started
progress: 80% of clients started
progress: 90% of clients started
progress: 100% of clients started
Application protocol: h2c
Warm-up phase is over for thread #0.
Main benchmark duration is started for thread #0.
=================================================================
==5160==ERROR: AddressSanitizer: heap-use-after-free on address 0x616000010110 at pc 0x000000557a4e bp 0x7f5b2f19c720 sp 0x7f5b2f19c718
READ of size 8 at 0x616000010110 thread T1
    #0 0x557a4d in h2load::(anonymous namespace)::warmup_timeout_cb(ev_loop*, ev_timer*, int) /pat/to/nghttp2/src/h2load.cc:299:5
```

It also crashes in the different location when I stop server during main benchmarking duration.
It also crashes when I start h2load while server is not running.

src/h2load.cc
sohamm17: We thought that separating out rate_based and timing_based would be good, because it minimizes the confusion for the future developers.
Also, we wanted to initialize duration_watcher as late as possible. Because, if warm_up fails for some reason, we may not need duration_watcher.
@htuch Any thoughts on this?

sohamm17:
@tatsuhiro-t Fixed the memory error.

src/h2load.cc
tatsuhiro-t: Separation is nice.  My point is that because libev does not provide any method to check a given watcher is initialized or not, it is good to always initialize it in constructor, and stop them in destructor.  Initialization in this way gives us the known state of watcher, otherwise it is undefined.  We can change timeout value using ev_timer_set unless timer is started.  It is true that if warm_up is failed we don't need duration_watcher.  It is also true that initializing 1 or 2 extra watcher in constructor does not cause any impact on performance.

sohamm17: Okay. Fair enough.

tatsuhiro-t:
Thank you!  The crash is gone now.

I should use -m100 option before, but I just did it now.  It causes assertion error:

```
$ src/h2load -c10 -D10 --warm-up-time=5 -n0 http://localhost:3000 -m100
starting benchmark...
spawning thread #0: 10 total client(s). Timing-based test with 5s of warm-up time and 10s of main duration for measurements.
Warm-up started for thread #0.
progress: 10% of clients started
progress: 20% of clients started
progress: 30% of clients started
progress: 40% of clients started
progress: 50% of clients started
progress: 60% of clients started
progress: 70% of clients started
progress: 80% of clients started
progress: 90% of clients started
progress: 100% of clients started
Application protocol: h2c
Warm-up phase is over for thread #0.
Main benchmark duration is started for thread #0.
h2load: h2load.cc:1391: void h2load::Worker::sample_req_stat(h2load::RequestStat *): Assertion `stats.req_stats.size() <= max_samples' failed.
```

sohamm17:
Thanks. Yes, I know this. We have not changed the `max_samples` that `h2load` can take, from what you have set i.e 1,000,000. If we generate more than 1,000,000 requests, then this assertion fails. The user needs to reduce either duration or number of clients or number of concurrent connections for this assertion to pass.

tatsuhiro-t:
Understood.  I'll handle this assertion error.  This PR looks good.
I'll merge this PR after v1.25.0 is released.  Thank you.

tatsuhiro-t:
Assertion failure issue can be fixed by switching from current systematic sampling to reservoir sampling.  Will be done in a separate PR.

tatsuhiro-t:
Thank you.  Merged now.

PR #971, opened 2017-08-01
Patch for libnghttp2_asio to support Visual Studio 2015.

pengweichu:
Author: Liu Yong Guan

### DIFF
--- a/src/HttpServer.cc
+++ b/src/HttpServer.cc
@@ -586,7 +586,7 @@ void Http2Handler::start_settings_timer() {
 
 int Http2Handler::fill_wb() {
   if (data_pending_) {
-    auto n = std::min(wb_.wleft(), data_pendinglen_);
+    auto n = (std::min)(wb_.wleft(), data_pendinglen_);
     wb_.write(data_pending_, n);
     if (n < data_pendinglen_) {
       data_pending_ += n;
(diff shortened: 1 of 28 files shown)


src/HttpServer.cc
```diff
@@ -586,7 +586,7 @@ void Http2Handler::start_settings_timer() {
 
 int Http2Handler::fill_wb() {
   if (data_pending_) {
-    auto n = std::min(wb_.wleft(), data_pendinglen_);
+    auto n = (std::min)(wb_.wleft(), data_pendinglen_);
```
tatsuhiro-t: I'm not sure why VS2015 doesn't work with std::min(a, b), and requires extra parentheses?  It makes many noises in this PR.

pengweichu: Since the MS defined the min and max macros so the std::min and std::max can't works with VS.
Please see this link: https://stackoverflow.com/questions/2789481/problem-calling-stdmax

tatsuhiro-t: Thank you.  Are there any other workaround?

pengweichu: No more stuffs, I just find two ways in google, one is add the extra parentheses, or defined `#define NOMINMAX `before including windows.h.

tatsuhiro-t: Is defining NOMINMAX less intrusive?

portsip: Maybe, but I prefer add extra parentheses -:)

tatsuhiro-t: Personally, I don't want to add extra parentheses.  It looks strange to me.

PR #983, opened 2017-08-12
lib: add nghttp2_rcbuf_is_static()

addaleax:
Add a `nghttp2_rcbuf_is_static()` method to tell whether a rcbuf
is statically allocated.

This can be useful for language bindings that wish to avoid
creating duplicate strings for these buffers; concretely, I am
planning to use this in the Node HTTP/2 module that is being
introduced.


tatsuhiro-t:
Thank you.  Merged now.

PR #988, opened 2017-08-16
Refactoring include directories

dvetutnev:
Refactoring include directories for build as CMake subdirectory (add_subdirectory(nghttp2))
Example: https://github.com/dvetutnev/tst_nghttp2


tatsuhiro-t:
Thank you.  Merged now.

PR #991, opened 2017-08-19
Fix unreachable links and improve Docker image

GitaiQAQ:
The docker is very friendly to us.

But the dockerfile was no longer working, libev and zlib links fail, c-ares are not included. 

So I made a new one.

Fix the link that is not available, and update the dependency, and export the version as an environment variable

The standalone-toolchain build work on the same layer, you can reduce the volume of pull, pre-download source code can use the docker cache to speed up.

Install all binaries and libraries under `sysroot` for easy viewing and management.

The information about this image is as follows:

  * https://hub.docker.com/r/gitai/nghttp2
  * [](https://microbadger.com/images/gitai/nghttp2-android-ci "Get your own image badge on microbadger.com")

I hope it is faster and more efficient, and helpful to everyone.

### DIFF
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -15,6 +15,7 @@
 FROM gitai/android-standalone-toolchain
 MAINTAINER Tatsuhiro Tsujikawa
 
+ENV PREFIX /root/usr/local
 
 # Setup version of libraries
 ENV OPENSSL_VERSION 1.0.2d


tatsuhiro-t:
Thank you.  In my opinion, it would be nicer to just fix build error without changing base image.

lubinson:
@GitaiQAQ 

Please help to update the image to new version or  manual to build image by ourselves. Is that run each build steps on a base image that will be many steps:)

@tatsuhiro-t 
As @GitaiQAQ said docker is close to us. Can nghttp2 make a project to move nghttp2 to cloud. Because as we know install nghttp2 is very difficult, especially for non-c users. And no yum or other integrate installation method then we can easilly make image by running yum install on a base image.  Please consider this. Thanks.

PR #1001, opened 2017-08-24
Fix OCSP related error when building with BoringSSL

rlei:
BoringSSL has no `openssl/ocsp.h` nor most OCSP related APIs used in
`shrpx_tls.cc`. This commit add ifdefs to disable related code to allow
building nghttp2 with BoringSSL (again).

It's possible to use `!defined(OPENSSL_IS_BORINGSSL)`, but since BoringSSL
defines `OPENSSL_NO_OCSP` which is more specific, I chose to go with the
latter one.


tatsuhiro-t:
Thank you!  Merged now.

PR #1002, opened 2017-08-26
Just fix unreachable links and remove ndk...

GitaiQAQ:
Just fix the link that is not available, and export the version as an environment variable.

The standalone-toolchain build work on the same layer, reduce the volume of pull.

Keep the base image and DockerFile structure.

And apologize for the last messy pr!

### DIFF
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -10,39 +10,47 @@
 #
 # $ sudo docker run -v /path/to/dest:/out nghttp2-android cp /root/build/nghttp2/src/nghttpx /out
 
-FROM ubuntu:vivid
 
+# Only use standalone-toolchain for reduce size
+FROM ubuntu:xenial
 MAINTAINER Tatsuhiro Tsujikawa
-
-ENV ANDROID_HOME /root/android
-ENV PREFIX $ANDROID_HOME/usr/local
+ENV ANDROID_HOME /root
 ENV TOOLCHAIN $ANDROID_HOME/toolchain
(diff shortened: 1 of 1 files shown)


Dockerfile.android
```diff
@@ -87,11 +95,11 @@ RUN patch -p1 < ../libev-4.19-android.patch && \
     make install
 
 WORKDIR /root/build
-RUN curl -L -O http://zlib.net/zlib-1.2.8.tar.gz && \
-    tar xf zlib-1.2.8.tar.gz && \
-    rm zlib-1.2.8.tar.gz
+RUN curl -L -O https://downloads.sourceforge.net/project/libpng/zlib/$ZLIB_VERSION/zlib-$ZLIB_VERSION.tar.gz && \
```
tatsuhiro-t: Could you tell me why zlib.net is not used?  I think it is official.

GitaiQAQ: Yes, this is the official version. But officials seem to not store older versions here.

When new versions are released (1.2.11), the old version (1.2.8) will be deleted.

Sourceforge.net is a official produced mirror and contains outdated versions.

Of course, also can be replaced with the official git repo to get this library.

However, if you know the storage location of the old versions on the zlib.net, I'm going to modify it.

tatsuhiro-t: Thank you.  I realized that zlib.net links to sf.net/libpng.  No problem.

tatsuhiro-t:
Thank you.  Merged now.

PR #1020, opened 2017-10-04
Reduce libxml2 version requirement to 2.6.26

FireBurn:
It's currently set at 2.7.7 but reducing this down to 2.6.26 allows nghttp2 to be built on RHEL5

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,7 +79,7 @@ else()
   set(ENABLE_PYTHON_BINDINGS_DEFAULT OFF)
 endif()
 
-find_package(LibXml2 2.7.7)
+find_package(LibXml2 2.6.26)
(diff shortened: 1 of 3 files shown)


tatsuhiro-t:
RHEL5 is too old to me.  Is it still supported by RH?
That said, version requirement is pretty much arbitrary; it was just based on Ubuntu LTS when I wrote the code.

FireBurn:
It's not still supported by Redhat, I still have a few machines that I can't get rid off at the moment but do have the ability to upgrade httpd and was bundling openssl and nghttp2. If it's an arbitrary number rather than based on features would you mind lowering it? It builds fine here

tatsuhiro-t:
I'm fine to lower the version requirement of libxml2.

PR #1029, opened 2017-10-18
Fixed accidental compiler flags concatenation for MSVC.

lazyhamster:
Your CMake script replaces /DNDEBUG compiler option with empty script.

But if CMAKE_C_FLAGS_{build} contains /DNDEBUG in the middle of other compiler switches then replacing it with empty string concatenates neighboring switches into one. And that make Visual Studio unhappy because it can not recognize that we have 2 switches and not 1.
In my case `/Oi /Gy /DNDEBUG /Z7` became `/Oi /Gy/Z7` and VS for some reason thinks that /Gy/Z7 is a single /GZ switch.

This patch eliminate the issue. In general extra space should not matter and in this particular case it will solve the problem with switches.


tatsuhiro-t:
Thank you!  Merged now.

PR #1063, opened 2017-11-19
Error callback2

### LINKED ISSUE #1062: Using *_set_error_callback to detect non-http2 peer
For the Node.js http2 implementation, we the following test case:

```js
'use strict';

const common = require('../common');
if (!common.hasCrypto)
  common.skip('missing crypto');

const http = require('http');
const http2 = require('http2');

const server = http.createServer(common.mustNotCall());

server.listen(0, common.mustCall(() => {
  const client = http2.connect(`http://localhost:${server.address().port}`);

  client.on('error', console.log);
  client.on('socketError', console.log);

  client.on('close', (...args) => {
    server.close();
  });

}));
```

In other words, an http2 client talking to an http1 server. When this happens, we want to be able to report an error to the `client` (e.g. we want `client.on('error')` to receive an appropriate error.

However, currently when nghttp2 is processing the invalid response from the http1 server, the best we get currently is a callback to the `*_set_error_callback` with a message that says, `Remote peer returned unexpected data while we expected SETTINGS frame.  Perhaps, peer does not support HTTP/2 properly."

While we can check for this condition fairly easily within the `*_set_error_callback`, doing so requires us to compare the value of the `msg`, which is less than ideal. An alternate solution would be to add a `*_set_error_callback2` variant that would receive a proper numeric error constant that we can key off rather than using the string value.

tatsuhiro-t:
Fixes #1062

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -81,6 +81,7 @@ APIDOCS= \
 	nghttp2_session_callbacks_set_before_frame_send_callback.rst \
 	nghttp2_session_callbacks_set_data_source_read_length_callback.rst \
 	nghttp2_session_callbacks_set_error_callback.rst \
+	nghttp2_session_callbacks_set_error_callback2.rst \
 	nghttp2_session_callbacks_set_on_begin_frame_callback.rst \
 	nghttp2_session_callbacks_set_on_begin_headers_callback.rst \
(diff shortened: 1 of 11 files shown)


jasnell:
Thank you @tatsuhiro-t !

PR #1120, opened 2018-01-27
Fix #1119: Overwrite of first header on mruby call to env.req.set_header(..)

### LINKED ISSUE #1119: nghttpx: In mruby, the first header is always deleted on every call to env.req.set_header(...)
A bug in the nghttpx mruby code has been stripping all of our HTTP/2 request headers when a certain mruby script was enabled. This script called `env.req.set_header("HEADER", "")` multiple (10+ times) in order to strip-away insecure headers before passing the request internally. The bug that we observed was that all headers were being stripped away, regardless of whether they were in a `set_header` directive or not.

The offending bug is in the `src/shrpx_mruby_module_request.cc` file, in the `request_mod_header` function, and only occurs when `repl == true`. This is due to a simple logic error in the header replacement code, which will always truncate the headers list by removing the first header.

A pull request with the fix is ready and will be attached momentarily.

dylanplecki:
Fixes #1119.

Fixes a simple logic bug where the first header will be overwritten when the replacement header is not equal to the first header in the header hashmap. In this scenario, after `p == i == 0`, then `p == (i - 1)`, which iteratively overwrites the previous value, starting with index 0 (the first header), then resizes the array to `size() - 1`.

This patch fixes the bug so that `p != i` iff at least one header key has matched the replacement header key.

### DIFF
--- a/src/shrpx_mruby_module_request.cc
+++ b/src/shrpx_mruby_module_request.cc
@@ -246,8 +246,9 @@ mrb_value request_mod_header(mrb_state *mrb, mrb_value self, bool repl) {
         continue;
       }
       if (i != p) {
-        headers[p++] = std::move(kv);
+        headers[p] = std::move(kv);
       }
+      ++p;
     }
     headers.resize(p);
   }


tatsuhiro-t:
Thank you for PR.  Yes, this is a real bug.

tatsuhiro-t:
The same bug exists in shrpx_mruby_module_response.cc.  I will deal with it.

PR #1132, opened 2018-02-17
 Allow building without NPN 

Sp1l:
NPN has been superseeded by ALPN. OpenSSL provides a configure
option to disable npn (no-npn) which results in an OpenSSL
installation that defines OPENSSL_NO_NEXTPROTONEG in opensslconf.h

The #ifdef's look safe here (as the next_proto is initialized as
nullptr). Alteratively, macros could be defined for the used npn
methods that return a 0 for next_proto.


tatsuhiro-t:
Thank you for PR.  It looks like the relevant commit is 2073566 only.  Could you rebase commits to include just that?  Also a callback function used in SSL_CTX_set_next_proto_select_cb (e.g., next_proto_cb in shrpx_tls.cc) should be surrounded by `#ifndef OPENSSL_NO_NEXTPROTONEG` and `#endif // !OPENSSL_NO_NEXTPROTONEG` too.

Sp1l:
Rebasing

PR #1141, opened 2018-03-05
Support Visual Studio

dvetutnev:
- Appveyor
Add build ZLIB from sources
Add build nghttp2_asio
- Travis
Add build Boost from sources (for compatibility with gcc-7)
Add build nghttp2_asio
Upgrade CMake (install binary from cmake.org)
- nghttp2_asio
Add DLL-interface (__declspec(dllexport)/__declspec(dllimport))
Add function, missing in MSVC (based on PR #971)
Use define NOMINMAX for compatibility std::min/std::max in MSVC
Use CMake imported targets for link nghttp2_asio and applications


tatsuhiro-t:

PR #1146, opened 2018-03-12
cmake: add ENABLE_STATIC_LIB option to build static lib

vszakats:
When using the `ENABLE_LIB_ONLY` option, only the shared library was built. This new option allows to build the static library as well.


vszakats:
I use this to create `curl` builds (https://github.com/vszakats/curl-for-win) for https://curl.haxx.se/download.html.

tatsuhiro-t:
Thank you. LGTM

PR #1152, opened 2018-03-25
[PATCH] Allow building without NPN

Sp1l:
NPN has been superseeded by ALPN. OpenSSL provides a configure
option to disable npn (no-npn) which results in an OpenSSL
installation that defines OPENSSL_NO_NEXTPROTONEG in opensslconf.h

The #ifdef's look safe here (as the next_proto is initialized as
nullptr). Alteratively, macros could be defined for the used npn
methods that return a 0 for next_proto.

Signed-off-by: Bernard Spil <brnrd@FreeBSD.org>


tatsuhiro-t:

PR #1153, opened 2018-03-25
Fix build with LibreSSL 2.7

Sp1l:
LibreSSL 2.7 added a large part of the OpenSSL 1.1 API.
 - Rename LIBRESSL_IN_USE to LIBRESSL_1_0_API (LibreSSL with
   OpenSSL 1.0 API)
 - Add LIBRESSL_1_1_API for ifdef blocks where the 1.1 API in
   LibreSSL is incomplete
 - Use OPENSSL_NO_PSK to detect missing PSK feature (not LibreSSL
   specific)

Signed-off-by: Bernard Spil <brnrd@FreeBSD.org>


src/shrpx_tls.cc
```diff
@@ -679,9 +679,9 @@ unsigned int psk_server_cb(SSL *ssl, const char *identity, unsigned char *psk,
   return static_cast<unsigned int>(secret.size());
 }
 } // namespace
-#endif // !LIBRESSL_IN_USE
+#endif // OPENSSL_NO_PSK
 
-#if !LIBRESSL_IN_USE
+#if OPENSSL_NO_PSK
```
tatsuhiro-t: I think this should be `#ifndef OPENSSL_NO_PSK`

tatsuhiro-t:
Thank you for PR.

I have no hard feelings toward LibreSSL, but supporting OpenSSL, LibreSSL, and boringSSL with slightly different/added/missing APIs is pretty much hard for me.  It is certainly not impossible, but.

Perhaps, the following criteria would make our lives easier:
For boringSSL, because it does not have proper release, we only care about its HEAD.
For OpenSSL, because it is our primary TLS stack, we support for its release versions.
For LibreSSL, it would be easier for us to just support its latest release.

tatsuhiro-t:
Another option is check availability of every OpenSSL API functions rather than checking library versions.  It forces me to tackle with cmake which I don't know much about.

tatsuhiro-t:
I did this differently in #1162.  I explicitly defined LIBRESSL_2_7_API instead.  This would be better because when libressl 2.8 comes out and it implements another set of OpenSSL 1.1 API, then we have to change LIBRESSL_1_1_API definition and add more ifdefs to exclude libressl 2.7.

tatsuhiro-t:
#1162 fixed libressl 2.7 issue.

PR #1161, opened 2018-04-11
Add attribute format in nghttp2_debug_vprintf_callback

ewwissi:
Change-Id: I991b84ce731adbd25f5cdf565333db1a19e75ec6


ewwissi:
this is for build warning in application which uses nghttp2_debug_vprintf_callback

PR #1184, opened 2018-05-31
Fix handling of SETTINGS_MAX_CONCURRENT_STREAMS.

PiotrSikora:
The maximum number of outgoing concurrent streams is initially
limited to 100 to avoid issues when the local endpoint submits
lots of requests before receiving initial SETTINGS frame from
the remote endpoint, since sending them at once to the remote
endpoint could lead to rejection of some of the requests.

This initial limit is overwritten with the value advertised in
SETTINGS_MAX_CONCURRENT_STREAMS setting by the remote endpoint,
but previously, it wasn't lifted if the remote endpoint didn't
advertise that setting (implying no limits), in which case the
limit of 100 was retained, even though it was never advertised
by the remote endpoint.

### DIFF
--- a/lib/nghttp2_session.c
+++ b/lib/nghttp2_session.c
@@ -4410,6 +4410,12 @@ int nghttp2_session_on_settings_received(nghttp2_session *session,
     return session_call_on_frame_received(session, frame);
   }
 
+  if (!session->remote_settings_received) {
+    session->remote_settings.max_concurrent_streams =
+        NGHTTP2_DEFAULT_MAX_CONCURRENT_STREAMS;
+    session->remote_settings_received = 1;
+  }
+
   for (i = 0; i < frame->settings.niv; ++i) {
     nghttp2_settings_entry *entry = &frame->settings.iv[i];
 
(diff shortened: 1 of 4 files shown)


lib/nghttp2_session.c
```diff
@@ -4410,6 +4410,12 @@ int nghttp2_session_on_settings_received(nghttp2_session *session,
     return session_call_on_frame_received(session, frame);
   }
 
+  if (!session->remote_settings_received) {
+    session->remote_settings.max_concurrent_streams =
+        NGHTTP2_DEFAULT_MAX_CONCURRENT_STREAMS;
+    session->remote_settings_received = 1;
+  }
```
tatsuhiro-t: I'm wondering setting max_concurrent_streams to max somewhere near https://github.com/nghttp2/nghttp2/blob/bfd4d8b052a594eb61b39d0ff260fbd19856763d/lib/nghttp2_session.c#L5421 might make the PR simpler because we don't have to add the new field to nghttp2_session.

tatsuhiro-t: My suggestion doesn't work when we see the some part of SETTINGS and before calling nghttp2_session_send.  I think your original PR looks good to me.

PiotrSikora: Yeah, I think that removing initial limit in `NGHTTP2_IB_READ_FIRST_SETTINGS` could open the floodgates in case of partial read of the `SETTINGS` frame from a peer that advertises `SETTINGS_MAX_CONCURRENT_STREAMS`.

Thanks for investigating!

tatsuhiro-t:

PR #1198, opened 2018-07-12
Fix compiling static lib on Windows with MSVC

sesmith177:
On Windows, the DLL will create nghttp2.lib AND nghttp2.dll. Then the
static lib also tries to build nghttp2.lib, but that fails because it
already exists.

The nghttp2.lib created for the DLL cannot be used without the DLL. This
compiles the static nghttp2.lib into a separate directory, which works
around the above issue.

Signed-off-by: Matthew Horan <mhoran@pivotal.io>


htuch:
Friendly ping, we'd like to unblock Envoy Windows build support with this. Thanks!

PR #1199, opened 2018-07-12
pulled default NGHTTP2_SETTINGS_MAX_CONCURRENT_STREAMS to 1073741824

prune998:
We need to get rid of the default 100 HTTP/2 streams as default... default should be unlimited (ideally ((1U << 31) - 1))


PiotrSikora:
The underlying issue has been fixed in #1184.

PR #1212, opened 2018-08-19
add option reuse addr in local endpoint configuration of asio client

akonskarm:
Hi Tatsuhiro

Adding reuse addr to get rid of annoying message when re-running same program shortly after closing it.

Best
Alexandros

### DIFF
--- a/src/asio_client_session_tcp_impl.cc
+++ b/src/asio_client_session_tcp_impl.cc
@@ -40,7 +40,11 @@ session_tcp_impl::session_tcp_impl(
     const std::string &host, const std::string &service,
     const boost::posix_time::time_duration &connect_timeout)
     : session_impl(io_service, connect_timeout),
-      socket_(io_service, local_endpoint) {}
+      socket_(io_service, tcp::v4()) {
+        boost::asio::socket_base::reuse_address option(true);
+        socket_.set_option(option);
+        socket_.bind(local_ep);
+}
 
 session_tcp_impl::~session_tcp_impl() {}
 


src/asio_client_session_tcp_impl.cc
```diff
@@ -40,7 +40,11 @@ session_tcp_impl::session_tcp_impl(
     const std::string &host, const std::string &service,
     const boost::posix_time::time_duration &connect_timeout)
     : session_impl(io_service, connect_timeout),
-      socket_(io_service, local_endpoint) {}
+      socket_(io_service, tcp::v4()) {
```
tatsuhiro-t: Is tcp::v4() required?

akonskarm: Hi Tatsuhiro

Unfortunately if we just call socket(io_service) the socket will be created in the closed state.
The socket needs to be in the open state to be able to set the option.
So it's a chicken and egg problem. We want to open on local_endpoint, but need to specify reuse address be fore being able to do so.

Maybe there is an alternative to opening the socket that I'm unaware of.

Best
Alexandros

src/asio_client_session_tcp_impl.cc
```diff
@@ -40,7 +40,11 @@ session_tcp_impl::session_tcp_impl(
...
+        boost::asio::socket_base::reuse_address option(true);
+        socket_.set_option(option);
+        socket_.bind(local_ep);
```
tatsuhiro-t: I think this should be local_endpoint

src/asio_client_session_tcp_impl.cc
tatsuhiro-t: Got it.  Thank you.

PR #1217, opened 2018-08-27
fix code for reuse addr on asio client

akonskarm:
Hi Tatsuhiro,

I revisited the asio client local endpoint code and I think it makes more sense to do it like this.
The previous code might be problematic if local_endpoint is ipv6.

Best regards,
Alexandros


tatsuhiro-t:
LGTM

PR #1222, opened 2018-09-05
Use __has_declspec_attribute for shared builds

donny-dont:
Clang compilation targets may support `__declspec` attributes. Because of that it has a `__has_declspec_attribute` [check](https://clang.llvm.org/docs/LanguageExtensions.html#has-declspec-attribute).

Currently nghttp2 just assumes that `__declspec` is Windows only. This adds a compatibility macro and checks whether `dllimport` and `dllexport` are available.

### DIFF
--- a/lib/includes/nghttp2/nghttp2.h
+++ b/lib/includes/nghttp2/nghttp2.h
@@ -31,6 +31,11 @@
 #  define WIN32
 #endif
 
+/* Compatibility for non-Clang compilers */
+#ifndef __has_declspec_attribute
+#  define __has_declspec_attribute(x) 0
+#endif
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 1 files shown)


donny-dont:
@tatsuhiro-t anything contentious here? Apologies for the nag its just been a bit for this PR.

tatsuhiro-t:
Is this macro evaluates to true only when building for Windows?

donny-dont:
@tatsuhiro-t its not necessarily a windows only thing. Other platforms using nghttp2 can use it.

See https://doc.dl.playstation.net/doc/ps4-oss/

tatsuhiro-t:
For those platforms, doesn't `__attribute__((visibility("default")))` work just like gcc?

donny-dont:
If it did I wouldn't be giving you this patch 😄

donny-dont:
@tatsuhiro-t any thoughts?

tatsuhiro-t:
Sorry for replying late.  LGTM.

Issue #1225, opened 2018-09-13
Can be static_table const?


Hadatko:
Hi,
currently it looks like this global variable is not const and it can be. Then linker can move it from rw section to ro section. I am right? Or are data in this global variable changed during communication. My very simple test looks fine when i changed it to const. File: **nghtt2_hd.c**

tatsuhiro-t:
Thank you.  It is now const.  See dfc0f248c60140194626c526221e45b6ee0b3d45

Hadatko:
Thank you too.

tatsuhiro-t:
I got compile errors with the suggested changes.

Hadatko:
Maybe you didn't change const everywhere. I change around 4 data types to const. But no other type of change. Maybe i can give a pull request.

PR #1229, opened 2018-09-14
Make const members which deserves it.

Hadatko:
Based on conversation #1225
I don't have compile errors. Do you?


tatsuhiro-t:
Yes.  See https://travis-ci.org/nghttp2/nghttp2/jobs/428606077

Hadatko:
I will postpone this. In my compiler it is working. i will try different one later.

PR #1232, opened 2018-09-26
adding an accessor for enabling CONNECT

alyssawilk:
As a follow-up to #1231, allowing setting NGHTTP2_SETTINGS_ENABLE_CONNECT_PROTOCOL via nghttp2_option_set_enable_connect_protocol.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -39,6 +39,7 @@ set(APIDOCS
   nghttp2_option_set_no_auto_ping_ack.rst
   nghttp2_option_set_no_auto_window_update.rst
   nghttp2_option_set_no_http_messaging.rst
+  nghttp2_option_set_enable_connect_protocol.rst
   nghttp2_option_set_no_recv_client_magic.rst
   nghttp2_option_set_peer_max_concurrent_streams.rst
   nghttp2_option_set_user_recv_extension_type.rst
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -64,6 +64,7 @@ APIDOCS= \
 	nghttp2_option_set_no_auto_window_update.rst \
 	nghttp2_option_set_no_closed_streams.rst \
 	nghttp2_option_set_no_http_messaging.rst \
+	nghttp2_option_set_enable_connect_protocol.rst \
 	nghttp2_option_set_no_recv_client_magic.rst \
 	nghttp2_option_set_peer_max_concurrent_streams.rst \
 	nghttp2_option_set_user_recv_extension_type.rst \
--- a/lib/includes/nghttp2/nghttp2.h
+++ b/lib/includes/nghttp2/nghttp2.h
@@ -2537,6 +2537,17 @@ NGHTTP2_EXTERN void nghttp2_option_set_no_http_messaging(nghttp2_option *option,
 /**
  * @function
  *
+ * By default, nghttp2 library does not allow the CONNECT method.  For
+ * applications which wish to use CONNECT to allow websockets to traverse an
+ * HTTP/2 hop, give a nonzero |val| to enable CONNECT support.
+ * See `RFC 8441 <https://tools.ietf.org/html/rfc8441>`_ for details.
+ */
+NGHTTP2_EXTERN void nghttp2_option_set_enable_connect_protocol(nghttp2_option *option,
+                                                         int val);
+
+/**
+ * @function
+ *
  * RFC 7540 does not enforce any limit on the number of incoming
  * reserved streams (in RFC 7540 terms, streams in reserved (remote)
  * state).  This only affects client side, since only server can push
--- a/lib/nghttp2_option.c
+++ b/lib/nghttp2_option.c
@@ -59,6 +59,11 @@ void nghttp2_option_set_no_http_messaging(nghttp2_option *option, int val) {
   option->no_http_messaging = val;
 }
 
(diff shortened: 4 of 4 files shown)


tatsuhiro-t:
CONNECT protocol just works by sending SETTINGS_ENABLE_CONNECT_PROTOCOL in SETTINGS frame. Any more options or disabling HTTP messaging is not required.

alyssawilk:
So what I am hearing you say is that I have a client wishing to do CONNECT to a server wishing to receive connect the client could could send SETTINGS_ENABLE_CONNECT_PROTOCOL followed by a header frame with CONNECT and it should work?

If change my local Envoy code to
iv.push_back({SETTINGS_ENABLE_CONNECT_PROTOCOL, 0});
immediately before 
nghttp2_submit_settings and then send a CONNECT request, the CONNECT is rejected by the server.

In my integration test I can see that the upstream server receives the settings frame (type 4) and sets session->remote_settings.enable_connectprotocol to 1.
The server then receives the headers, and calls http_request_on_header with 
session->local_settings.enable_connectprotocol which is 0

When the :protocol NGHTTP2_TOKEN__PROTOCOL header is received by the server, connect_protocol is then false in http_request_on_header, and NGHTTP2_ERR_HTTP_HEADER is returned.

If you believe that connect support should be enabled by the peer sending the correct settings frame, perhaps http_request_on_header should be validating based on 
session->remote_settings.enable_connectprotocol?  I would be happy to amend this PR if that is the case.  Alternately if there is something else wrong with the http2 flow I am describing I would appreciate a pointer as to what I am doing wrong.

tatsuhiro-t:
I believe that SETTINGS_ENABLE_CONNECT_PROTOCOL must be sent by server, not client.
And server validates a request including :protocol using its local settings.

From RFC 8441 https://tools.ietf.org/html/rfc8441#section-3:

>    Upon receipt of SETTINGS_ENABLE_CONNECT_PROTOCOL with a value of 1, a
   client MAY use the Extended CONNECT as defined in this document when
   creating new streams.  Receipt of this parameter by a server does not
   have any impact.

alyssawilk:
If a server validates request using its local settings, why is it not OK for the server to set that option via accessor, i.e. 

nghttp2_option_set_enable_connect_protocol(options_, 1)?

Envoy currently roughly
nghttp2_option_new(&options_);
nghttp2_option_set_foo() // for all various optins
nghttp2_session_server_new2(.... options_)

so I'm trying to extend that to allow setting nghttp2_option_set_enable_connect_protocol(options_, 1); to be consistent with all our other local options.

tatsuhiro-t:
nghttp2_option_set_* is offered for behaviour which cannot change using nghttp2_submit_settings.
If feature can be enabled by nghttp2_submit_settings, we don't offer the same functionality in nghttp2_option_set_*.
I think Envoy server can just send NGHTTP2_SETTINGS_ENABLE_CONNECT_PROTOCOL = 1 in SETTINGS and it should enjoy RFC 8441 just fine.

alyssawilk:
My envoy PR has the server sending NGHTTP2_SETTINGS_ENABLE_CONNECT_PROTOCOL = 1 .  The server is still rejecting incoming :protocol headers.

As I look at nghttp2_session.c the only place where session->local_settings.enable_connect_protocol is set is in nghttp2_session_update_local_settings, which appears to not be called when the server executes nghttp2_submit_settings, it is called but when settings are *received* with an NGHTTP2_FLAG_ACK

tatsuhiro-t:
session->pending_enable_connect_protocol is used to check whether :protocol is allowed on server side.  It becomes 1 when SETTINGS_ENABLE_CONNECT_PROTOCOL is passed to nghttp2_submit_settings.

alyssawilk:
Ah, it does now.  That code was submitted https://github.com/nghttp2/nghttp2/commit/651e14771182e2292833d84ab513e5f018df51c2, the day after I sent this PR out.
Thank you for fixing the issue I was encountering!

One further question, which does not relate to actually turning connect up, I notice you reject content-length on the websocket response.  In https://tools.ietf.org/html/rfc8441 I only see a notice "
Other HTTP request and
   response-header fields, such as those for manipulating cookies, may
   be included in the HEADERS with the CONNECT method as usual."
which no references to content length.  May I ask why you explicitly check and disallow content length?

tatsuhiro-t:
https://tools.ietf.org/html/rfc7231#section-4.3.6 says server must not send Content-Length in response to  the successful response to CONNECT method.  RFC says client must ignore them.  I think that those sever which sends Content-Length in CONNECT method is broken, and there is no harm to drop the stream.

alyssawilk:
Ah, I see, I was looking a the upgrade specs not the connect spec.

As I understand it, 101 response is allowed to have a body.  So if we are to strictly abide by the spec, the websocket upgrade response body would not survive the h2 hop and so would lose data.  I would prefer to allow users to preserve that data on the upgrade-downgrade path but I understand if you would prefer to strictly adhere to the CONNECT spec.

tatsuhiro-t:
According to https://tools.ietf.org/html/rfc7230#section-3.3.3 any response with 1xx cannot have message body.  Also https://tools.ietf.org/html/rfc7230#section-3.3.2 says that server must not send content-length in response with 1xx.  I think there is no risk to lose data in any h2/h1 mixed hop at least in response chain.  HTTP/2 stream can send any amount of data without content-length, and for h1 path, after empty line of 1xx response, it is not h1 anymore, and it goes to the end of the connection.

alyssawilk:
Oh fantastic, thanks for the link.

PR #1256, opened 2018-11-11
h2load: add an option to write per-request logs

pyos:
Current format is tab-separated (start timestamp, response code, duration in microseconds), with room for extension by adding more columns. Useful for custom aggregation, e.g. computing percentiles or selecting specific time ranges. The log has thread-local buffers of up to 16 lines, so performance impact is practically nonexistent, but rows can appear slightly out of order.

### DIFF
--- a/src/h2load.cc
+++ b/src/h2load.cc
@@ -748,6 +748,7 @@ void Client::on_header(int32_t stream_id, const uint8_t *name, size_t namelen,
       }
     }
 
+    stream.req_stat.status = status;
     if (status >= 200 && status < 300) {
       ++worker->stats.status[2];
       stream.status_success = 1;
@@ -775,6 +776,7 @@ void Client::on_status_code(int32_t stream_id, uint16_t status) {
     return;
   }
 
+  stream.req_stat.status = status;
   if (status >= 200 && status < 300) {
     ++worker->stats.status[2];
     stream.status_success = 1;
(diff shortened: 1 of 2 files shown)


roboslone:
@tatsuhiro-t could you take a look at this, please?

tatsuhiro-t:
Will do.  Thank you for PR.

src/h2load.cc
```diff
@@ -821,6 +823,22 @@ void Client::on_stream_close(int32_t stream_id, bool success, bool final) {
     }
     ++worker->stats.req_done;
     ++req_done;
+
+    if (worker->log) {
+      auto to_us = [](auto d) {
+        return std::chrono::duration_cast<std::chrono::microseconds>(d);
+      };
+      auto start = to_us(req_stat->request_wall_time.time_since_epoch());
+      auto delta = to_us(req_stat->stream_close_time - req_stat->request_time);
```
tatsuhiro-t: I'm not sure these kind of lambda is optimized, but it might be easier to just write duration_cast for each line.

src/h2load.cc
```diff
@@ -821,6 +823,22 @@ void Client::on_stream_close(int32_t stream_id, bool success, bool final) {
...
+      *worker->log << start.count() << '\t' << (success ? req_stat->status : -1)
+                   << '\t' << delta.count() << '\n';
+      // Flushing manually is important to ensure atomicity of lines, but
+      // doing it after each line (e.g. with std::endl) is noticeably slow.
+      if (++worker->log_pending == 16) {
+        worker->log->flush();
+        worker->log_pending = 0;
+      }
```
tatsuhiro-t: I'm wondering write(2) might help atomicity and performance.  C++ stream is a bit slow.

doc/h2load.1.rst
```diff
@@ -229,6 +229,16 @@ OPTIONS
...
+.. option:: --log-file=<FILE>
+
+    Write per-request information to a file as tab-separated
+    columns:  start time  as microseconds since epoch;  HTTP
+    status code;  microseconds  until end of response.  More
+    columns may be added later.  Rows are ordered by end-of-
+    response time  when using  one  worker  thread,  but may
+    appear  slightly  out  of order  with  multiple  threads
+    due to buffering.  Status code is -1 for failed streams.
+
```
tatsuhiro-t: These manual pages will be generated in some point in a batch using --help option, and it is not required in PR.

pyos: OK, I've removed this commit from the branch.

src/h2load.cc
pyos: Using `write` here would also require something like `snprintf` to format the entry, though, which iirc is actually slower than iostreams due to the overhead of parsing the format string.

src/h2load.cc
pyos: OK, I've inlined the lambda.

tatsuhiro-t:
Thank you.  Cherry-picked as de4fe72861d39396c1b5658c37ec200cdea2eb4a

Issue #1252, opened 2018-11-06
Automatic Port on asio_http2 server


promgamer:
Hi,

Are you interested in a automatic port attribution when using asio_http2 server? Right now, we need to tell the server a port to listen. But asio supports not defining a port (and ask the operating system for a random available port).  This allows one to start a server without defining a port (asio will search for the next available port) and have a function that returns what port is it listening.

I think this could be a nice addition. If you're ok with this, I can send a PR.

Let me know.

tatsuhiro-t:
Closed since asio library has been removed.

PR #1257, opened 2018-11-12
added access to the number of the current server port

puscas:
Adds support for the requested feature #1252

### DIFF
--- a/src/asio_server_http2_impl.cc
+++ b/src/asio_server_http2_impl.cc
@@ -78,6 +78,11 @@ http2_impl::io_services() const {
   return server_->io_services();
 }
 
+int http2_impl::port() const {
+  const auto ports = server_->ports();
+  return ports.empty() ? 0 : server_->ports().back();
+}
+
 } // namespace server
 
 } // namespace asio_http2
--- a/src/asio_server_http2_impl.cc
+++ b/src/asio_server_http2_impl.cc
@@ -78,9 +78,8 @@ http2_impl::io_services() const {
   return server_->io_services();
 }
 
-int http2_impl::port() const {
-  const auto ports = server_->ports();
-  return ports.empty() ? 0 : server_->ports().back();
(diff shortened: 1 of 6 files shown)


src/asio_server_http2_impl.cc
```diff
@@ -78,6 +78,11 @@ http2_impl::io_services() const {
   return server_->io_services();
 }
 
+int http2_impl::port() const {
+  const auto ports = server_->ports();
+  return ports.empty() ? 0 : server_->ports().back();
```
tatsuhiro-t: Is there any reason to just return a single port?  If it is enough to return just 1, it is not necessary to make a temporary vector.

tatsuhiro-t:
Thank you for PR.  Could you check my comment?

src/asio_server_http2_impl.cc
puscas: Inside the asio_server exists a vector for the acceptors, on each acceptor exists a port. In the tests I made the ASIO server creates 2 acceptors inside the vector, the first one have port 0 and the second one have the port I set. I'm not sure if more than two acceptors can be inside the vector in the nghttp ecosystem, so the vector with ports cope well with it.
However, if only one port is in use and the acceptors vector only have two acceptors, where the second one is the one with the open port number, your suggestion is the correct way to do it.

tatsuhiro-t: If you have two addresses resolved, you have 2 acceptors (e.g., 127.0.0.1 and ::1).
I'm not sure why you get port 0.  In my test, I got the same pot number for 2 acceptors.

puscas: You are right. I've made an error and that's why I received always zero in the first acceptor. Also I changed the code to always use a vector. Thanks for the feedback!

tatsuhiro-t:
Thank you.  Cherry-picked as eb6792533d0750edcfd431cafaa6740e026b37df

PR #1269, opened 2018-12-02
Fix compilation with gcc5, stdc++14

iMichka:
Fixes:
shrpx_api_downstream_connection.cc:57:3: error: array must be initialized with a brace-enclosed initializer


tatsuhiro-t:
I think compiling with gcc-6 fixes the compile error.

iMichka:
I can test that. If this is the case, is gcc-6 a hard requirement to build nghttp2?

PR #1270, opened 2018-12-02
Fix for compilation against modern LibreSSL

baitisj:


GallaFrancesco:
Is it possible to have an ETA on this? Looks like a minor fix, but it is necessary to build (and therefore use) nghttp2 with any version of libressl >= 2.8.1. Thank you!

tatsuhiro-t:
Sorry for the late reply,  merged now.  Thank you for PR.

PR #1280, opened 2018-12-23
Fix boost asio async object lifetimes

pacevedom:
While using boost::asio::async* functions we have to take special care when dealing with objects lifetime, since lots of callbacks are used for later execution.
In the event of closing connections before we have finished dealing with the messages (and writing back the responses) we can see some exceptions pop up because of already deleted connection objects. This is due to the capture list of lambda callbacks, that were created when the object was alive, and somewhere in between those calls and the actual ones that happen in asio, the connection is gone and the data members too.
The exceptions that we can see are std::bad_weak_ptr instaces, which happen whenever we need to create a new weak_ptr from an already freed (reset) shared_ptr, which is what shared_from_this enables.
This PR fixes that by adding some sanity checks at the beginning of the lambda functions, to bail out if the owning object has already been deleted.

The previous behavior can be tested by using gist https://gist.github.com/tatsuhiro-t/ba3f7d72d037027ae47b and then removing the sleep at line 72. Then, start sending requests with h2load with option -D and a high number of connections+streams, for example: -D 10 -c 300 -m 300. This issue will start to appear once time is up, since h2load will close the connection before letting our server finish off. Then, when responses try to write in the socket, it is already closed, the object is freed, and callbacks start to make use of freed memory.

### DIFF
--- a/src/asio_server_connection.h
+++ b/src/asio_server_connection.h
@@ -41,15 +41,15 @@
 
 #include <memory>
 
-#include <boost/noncopyable.hpp>
 #include <boost/array.hpp>
+#include <boost/noncopyable.hpp>
 
 #include <nghttp2/asio_http2_server.h>
 
 #include "asio_server_http2_handler.h"
 #include "asio_server_serve_mux.h"
-#include "util.h"
 #include "template.h"
+#include "util.h"
 
 namespace nghttp2 {
 
@@ -69,13 +69,10 @@ public:
       const boost::posix_time::time_duration &tls_handshake_timeout,
       const boost::posix_time::time_duration &read_timeout,
       SocketArgs &&... args)
-      : socket_(std::forward<SocketArgs>(args)...),
-        mux_(mux),
+      : socket_(std::forward<SocketArgs>(args)...), mux_(mux),
         deadline_(socket_.get_io_service()),
         tls_handshake_timeout_(tls_handshake_timeout),
-        read_timeout_(read_timeout),
-        writing_(false),
-        stopped_(false) {}
+        read_timeout_(read_timeout), writing_(false), stopped_(false) {}
 
   /// Start the first asynchronous operation for the connection.
   void start() {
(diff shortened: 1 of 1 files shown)


noosc:
meet the same exception, this really works for me, thanks

noosc:
> meet the same exception, this really works for me, thanks

but handler_ should take weak_ptr of connection, not shared_ptr

tatsuhiro-t:
It looks like just merging #1349 fixes this issue (other lambdas take std::shared_ptr so connection object is retained).  What do you think?

minhkstn:
Hi @noosc 
I got the same exception but your solution in #1349 does not work for me.

I have a client that requests multiple files sequentially. The server works well until the downloading time reaches around 108s.  At that time, the server stops sending any data although it has not finished handling the request.

After a period of time (e.g., 60s), the server finishes sending the data but it returns 
`terminate called after throwing an instance of 'std::bad_weak_ptr'
  what():  bad_weak_ptr
Aborted (core dumped)`
And the function `on_stream_close_callback()` in the client is not called.

In the example below, after sending some data of the file `No. 65` that starts at 108526ms, the server stops and joins again at time 169543ms.

**Here's the client side:**

 ```
recv (stream_id=51) :method: GET
 recv (stream_id=51) :scheme: http
 recv (stream_id=51) :authority: 172.16.1.1:3002
 recv (stream_id=51) :path: /file_64
 recv (stream_id=128) :status: 200
 recv (stream_id=128) date: Tue, 01 Sep 2020 19:40:51 GMT

[Minh - INFO] on_stream_close_callback() is called.

 recv (stream_id=51) :method: GET
 recv (stream_id=51) :scheme: http
 recv (stream_id=51) :authority: 172.16.1.1:3002
 recv (stream_id=51) :path: /file_65
 recv (stream_id=130) :status: 200
 recv (stream_id=130) date: Tue, 01 Sep 2020 19:40:54 GMT
Some requests were not processed. total=85, processed=83
```

**Here's the server side**
```
Respond file No. 64 at time: 105138ms
Sent file No. 64 at time: 108526ms

Respond file No. 65 at time: 108526ms
Sent file No. 65 at time: 169543ms

terminate called after throwing an instance of 'std::bad_weak_ptr'
  what():  bad_weak_ptr
Aborted (core dumped)
```

Do you have any idea to fix this issue?
Thanks

tatsuhiro-t:

PR #1282, opened 2018-12-26
Travis: Update to Xenial

alagoutte:
It is now possible to use Xenial (Ubuntu 16.04) on Travis

Also no longer need to get external clang (on Travis Xenial, it is clang-7)


tatsuhiro-t:
Thank you for PR!  Merged now.

Issue #373, opened 2015-09-28
Support BoringSSL


tatsuhiro-t:
Since we support OpenSSL and LibreSSL, it would be nice to also support BoringSSL.
Ref https://github.com/tatsuhiro-t/nghttp2/issues/291#issuecomment-143711259

bagder:
Let me just mention that BoringSSL is much harder to support than libressl and they more or less say that it is a moving target that isn't for non-google Software...

tatsuhiro-t:
Thank you.  I agree that moving target is really hard job to keep up especially for API changes.
I just imagined the differences are just  few configuration APIs, but it appears more substantial?

0x64616E69656C:
I agree with the argument of boringssl being a moving target, and should not replace openssl as default crypto-library. I have forked bssl internally for "personal use", modified the build system, kicked out a few obsolete/unneccessary/insecure ciphers, and merge changes from upstream boringssl to the remaining code. As far as implementation goes, boringssl uses openssl header files and provides -lcrypto and -lssl like openssl does, the issue with boringssl is that they dropped support for *_ctrl macros, as far as I understood, replacing those in nghttp2 with their proper functions should not affect compilation against openssl/libressl.

```
Switch any `*_ctrl` callers to the macro/function versions. This works in both
OpenSSL and BoringSSL. Note that BoringSSL's function versions will be
type-checked and may require more care with types
```

Which other issues are to be expected when providing support for boringssl that I might be missing?

tatsuhiro-t:
f0d2c9f94bbc08b22633f1a20a9ee1982c50d208 enables compilation with BoringSSL, except for neverbleed and libnghttp2_asio.
The latter  uses boost::asio, which calls OpenSSL functions deleted in BoringSSL.

BenPope:
It's a shame, because the patch to asio is small and significantly improves scalability across threads:
https://konradzemek.com/2015/08/16/asio-ssl-and-scalability/
https://gist.github.com/kzemek/37aa2a2138b2651f2c55

tatsuhiro-t:
Do you have a plan to submit the patch to boost upstream?

BenPope:
I think it's been done:
https://github.com/chriskohlhoff/asio/issues/52
https://github.com/chriskohlhoff/asio/issues/74
But my understanding is that Chris Kohlhoff is pretty busy with standardisation efforts.

tatsuhiro-t:
[...]

PR #1295, opened 2019-01-18
Fix compilation with boringssl

bratkartoffel:
Currently nghttp2 cannot be built against boringssl.
This patch adds some additional precompiler checks to choose the right api calls so the compilation against current boringssl (32e59d2d3264e4e104b355ef73663b8b79ac4093) succeeds.

Follow up to #373


alagoutte:
May be a good idea to add boringssl to (Travis) build system ?

PR #1303, opened 2019-02-06
Explicitly set install location when building shared libs

donny-dont:
Currently when building on Windows the dll and lib files are in the same location. This explicitly states where to install each part of the shared library.


tatsuhiro-t:
Thank you for PR.  Sorry for late response.

PR #1306, opened 2019-02-13
Fix build failing with make

lucifercr07:
Issue:
Error while building with make: array must be initialized with a brace enclosed initializer

Fix:
Enclosing array initialisation with proper braces.


jeroen:
+1

Kyle-Falconer:
I came up with this same change and tested that it works in Ubuntu 16.04. This change would resolve #1345

jay:
+1

Issue #1202, opened 2018-07-19
config.h


Hadatko:
Hi,
i wanted ask about 
#ifdef HAVE_CONFIG_H
#  include <config.h>
#endif /* HAVE_CONFIG_H */

shouldn't be used 
#  include "config.h"
instead of
#  include <config.h>
?

And even better could be renamed config.h to something more specific like nghttp2_config.h.

sarenameas:
I agree with this. Renaming this nghttp2_config.h helps with portability in an environment where there are possibly multiple config files.

Hadatko:
I made this pull request: https://github.com/nghttp2/nghttp2/pull/1311

Hadatko:
Looks like still an issue

PR #1311, opened 2019-02-21
#1202 Replaced config.h with more unique name nghttp_user_config.h

Hadatko:
-- Merged two solution into one (use everywhere nghttp2_config.h).
-- replace config.h with nghttp2_user_config.h

Hello, This is kindly request to replace config.h file name with some less common name, more specific to your project. Otherwise i have a conflict due to combining with other github projects which are using same file name inside their projects.

### DIFF
--- a/lib/nghttp2_buf.h
+++ b/lib/nghttp2_buf.h
@@ -25,9 +25,7 @@
 #ifndef NGHTTP2_BUF_H
 #define NGHTTP2_BUF_H
 
-#ifdef HAVE_CONFIG_H
-#  include <config.h>
-#endif /* HAVE_CONFIG_H */
+#include "nghttp2_config.h"
 
 #include <nghttp2/nghttp2.h>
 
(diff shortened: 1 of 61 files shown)


lib/nghttp2_buf.h
```diff
@@ -25,9 +25,7 @@
 #ifndef NGHTTP2_BUF_H
 #define NGHTTP2_BUF_H
 
-#ifdef HAVE_CONFIG_H
-#  include <config.h>
-#endif /* HAVE_CONFIG_H */
+#include "nghttp2_config.h"
```
sarenameas: I like better the pattern mbedtls has: 
```
#ifdef NGHTTP2_USER_CONFIG
#    include "nghttp2/config.h"
#endif
```
Keeping the compiler directives means we don't have to have a config file or run cmake. What do you think?

Hadatko: Then i will have same issue as with mbedtls :D

Hadatko: https://github.com/ARMmbed/mbedtls/pull/2459

PR #1313, opened 2019-03-03
nghttp2::asio_nghttp2::client. Add callbacks to session constructor

pacevedom:
This PR adds on_connect and on_error callbacks to nghttp2::asio_http2::client session objects. The goal is to avoid race conditions upon connections being successful or not.
The current implementation separates the session object connection (which is performed in the constructor) from the callbacks that are going to be executed when the connection is successful or not. This creates a situation where it might happen that the connection is established (or not), and callbacks are still not present, rendering the object in an unknown state to the caller.
Connection timeouts do not work here either, because those are for timeouts and this situation happens even in successful connections, or rejected ones.
The PR adds the possibility of setting up the callbacks in the constructor, making use of the RAII pattern. This way, there is no situation where the callbacks are not going to be called because they are set even before the connection is initiated.

### DIFF
--- a/src/asio_client_session.cc
+++ b/src/asio_client_session.cc
@@ -54,6 +54,27 @@ session::session(boost::asio::io_service &io_service,
 }
 
 session::session(boost::asio::io_service &io_service, const std::string &host,
+                 const std::string &service, connect_cb ccb, error_cb ecb)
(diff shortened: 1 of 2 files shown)


tatsuhiro-t:

PR #1349, opened 2019-05-24
fix bad_weak_ptr

noosc:
pass weak_ptr of connection to http2_handler, avoid calling `do_write` after connection is destroyed.
#1280


tatsuhiro-t:

PR #1365, opened 2019-06-24
Add max_concurrent_streams configuration to asio server

hap2712:
Enable configuration for max_concurrent_streams parameter to nghttp2::asio_http2, since it is hardcoded to 100.

Done through the nghttp2::asio_http2::server interface, defaults to 100 if not set.


tatsuhiro-t:

PR #1377, opened 2019-07-29
Support building nghttpx with systemd using cmake

Aldrog:


tatsuhiro-t:

PR #1394, opened 2019-09-16
Avoid filename collision of static and dynamic lib

wrowe:
Renames the output of the ENABLE_STATIC_LIB library/archive output
to nghttp2_static.lib/.a to avoid filenames colliding with the output
name for ENABLE_SHARED_LIB library/archive, when both are enabled.

This change requires a release notes notice.

Signed-off-by: William A Rowe Jr <wrowe@pivotal.io>
Signed-off-by: Yechiel Kalmenson <ykalmenson@pivotal.io>

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -62,7 +62,7 @@ if(HAVE_CUNIT OR ENABLE_STATIC_LIB)
   set_target_properties(nghttp2_static PROPERTIES
     COMPILE_FLAGS "${WARNCFLAGS}"
     VERSION ${LT_VERSION} SOVERSION ${LT_SOVERSION}
-    ARCHIVE_OUTPUT_NAME nghttp2
+    ARCHIVE_OUTPUT_NAME nghttp2_static
   )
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Thank you for PR. Merged now.

vszakats:
It means that now dependent projects using the statis libs are suddenly broken. I'd argue it's not good practice to use different names for dynamic and static libs, because dependent projects are almost always unable to adapt to such naming peculiarities.

However, there exists a standard to select static vs. shared e.g. in the GNU toolchain, so I'm guessing this is meant to fix the situation on Windows and MSVC specifically. In the meantime though, it just broke all other platforms. E.g. now I'll need to figure out a patch + nghttp2-specific configuration options for curl to pick the new names, which will require a custom patch, which will in turn break building with all preceding nghttp2 versions.

I think for MSVC users it may be a solution to run the build twice if both the static and dynamic results are required.

vszakats:
Issued a PR which makes the hard-coded lib name suffix optional/configurable.

PR #1395, opened 2019-09-16
Correct ssize_t on Windows

wrowe:
There is no standard C type ssize_t. The modern C99 introduces ptrdiff_t
which has an effectively equivilant (but subtly different) definition.

This patch makes ssize_t the equivilant byte-width of size_t on Windows,
using the stdint.h definition.

The remaining issue is that every consumer on Windows must continue to
define (either -Dssize_t=int64_t or -Dssize_t=ptrdiff_t) an override
because ssize_t was resolved in the nghttp2-private config.h. These uses
of ssize_t bleed out to the public API in nghttp2.h, leaving non-POSIX
environments like Windows with an undefined ssize_t type.

The cleanest solution in any later refactoring will be to switch all
external declarations using ssize_t to use the C99-standard ptrdiff_t.

This change requires a release notes notice as it causes an ABI change,
only for those on a 64-bit win32 build environment.

Signed-off-by: William A Rowe Jr wrowe@pivotal.io
Signed-off-by: Yechiel Kalmenson ykalmenson@pivotal.io

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -275,7 +275,11 @@ check_type_size("ssize_t" SIZEOF_SSIZE_T)
 if(SIZEOF_SSIZE_T STREQUAL "")
   # ssize_t is a signed type in POSIX storing at least -1.
   # Set it to "int" to match the behavior of AC_TYPE_SSIZE_T (autotools).
-  set(ssize_t int)
+  if(WIN32 AND CMAKE_SIZEOF_VOID_P EQUAL 8)
+    set(ssize_t int64_t)
+  else()
+    set(ssize_t int)
+  endif()
 endif()
 # AC_TYPE_UINT8_T
 # AC_TYPE_UINT16_T


tatsuhiro-t:
Thank you for PR.  I once tried to define ssize_t as SSIZE_T but failed because it breaks existing builds.
I think doing this by default is too late and it would be nice to add new cmake option to change the ssize_t definition for Windows.

wrowe:
> Thank you for PR. I once tried to define ssize_t as SSIZE_T but failed because it breaks existing builds.
> I think doing this by default is too late and it would be nice to add new cmake option to change the ssize_t definition for Windows.

Yes, SSIZE_T wouldn't work, for one it used to be the wrong bytewidth, plus we never want to be including "windows.h", but stick with standard C++ on both linux and windows. Fighting with the same chaos in several dependencies of Envoy Proxy.

Using ptrdiff_t is portable to windows and to unix because it is std c99. But that has to wait till you decide to do a 2.x binary break (the switch will be transparent to most compilations.)

In the public nghttp2.h, as a short term workaround, do we want to do a conditional;

#if defined (_MSC_VER) && !defined(ssize_t)
#include <stddef.h>
#define ssize_t ptrdiff_t
#endif

For any user this causes breakage for their local workaround, the workaround of the workaround would be -Dssize_t=ssize_t - passing the if define() test and keeping the token name.

Thoughts?

tatsuhiro-t:
I mean changing ssize_t to other than int in Windows build is not possible at the moment because the existing projects depend on it.

> For any user this causes breakage for their local workaround, the workaround of the workaround would be -Dssize_t=ssize_t - passing the if define() test and keeping the token name.

I think the default must be ssize_t to keep compatibility.  If users want to change ssize_t to other types (ptrdiff_t etc), they should do it explicitly by passing, for example, -Dssize_t=ptrdiff_t.

wrowe:
Could you tag this as a rev 1.0 version-major enhancement? I'm still looking at how to offer an explicit override with CMAKE, but changing all ssize_t -> ptrdiff_t throughout on a 1.0 release would be very helpful to non-posix consumers.

wrowe:
So as a top-level comment, if you would like to defer uio.h vs winsock2.h inclusion changes, the submission next week is going to propose we re-localize *all* platform network API includes into platform.h, where we can make the various envoy-centric tweaks (undefines on windows) in a single place.

tatsuhiro-t:
We have now nghttp2_ssize and its corresponding APIs and one can completely get rid of ssize_t if wanted.

PR #1411, opened 2019-10-29
priority_spec::valid(): remove const qualifier from return value

richard78917:
gcc generates warning:
* type qualifiers ignored on function return type [-Wignored-qualifiers]


tatsuhiro-t:
Thank you.  Merged now.

PR #1418, opened 2019-11-15
lib/CMakeLists.txt: Make hard-coded static lib suffix optional

vszakats:
It can be set via the `STATIC_LIB_SUFFIX` variable.

This fixes every existing dependent project that relied on the name
having no suffix and thus capable of using either a static or shared
flavour depending on which one is present on this or how the linker
is configured.

Ref: https://github.com/nghttp2/nghttp2/pull/1394

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -62,7 +62,7 @@ if(HAVE_CUNIT OR ENABLE_STATIC_LIB)
   set_target_properties(nghttp2_static PROPERTIES
     COMPILE_FLAGS "${WARNCFLAGS}"
     VERSION ${LT_VERSION} SOVERSION ${LT_SOVERSION}
-    ARCHIVE_OUTPUT_NAME nghttp2_static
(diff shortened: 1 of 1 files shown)


vszakats:
Any thoughts/comments?

vszakats:
Actually the Travis CI build appears to have finished [successfully](https://travis-ci.org/nghttp2/nghttp2/builds/612573650), not sure why it's still yellow here after 3 weeks.

If there is no hope for dealing with this, I'll abandon further [Windows builds for curl](https://github.com/curl/curl-for-win/) as they are now unmaintainable with the breakage introduced with 1.40.0. Shipping with an obsolete nghttp2 version isn't something tenable as well.

/cc @wrowe @tatsuhiro-t

tatsuhiro-t:
Sorry for the late relay, I have other things to do these days.  I think the change is good.
The CI is sometimes stuck somewhere and GitHub seems not to get webhook.  Not sure why.
Will merge soon.

vszakats:
Thank you @tatsuhiro-t!

Issue #1406, opened 2019-10-22
How to access SSL connection info with the C++ bindings


dke:
I am playing around with creating a HTTP2 client tool based on the `asio-cl2.cc` example, thus, based on the nghttp2 C++ bindings.

I struggle how to find out how to access underlying SSL stuff e.g. for printing SSL info (protocol name, cipher, tmp key info).

I found in the main project readme that the h2load program provides such output, but when I tried to learn there how to do it, it seems that the h2load program does not use the "automatic" SSL handling included in the C++ bindings used by the asio-cl2.cc example. I see a lot of manual SSL / networking code there.

So the question boils down to "how to access the `SSL*` pointer" hidden somewhere deep down below in the nghttp2 session object (or wherever else...), in order to be able to call something like
`SSL_get_current_cipher(const SSL *s);`.

I was expecting it somewhere in the `on_connect` callback, but I get lost browsing the code...

dke:
OK, I have done some research:

TL;DR: I guess as of now it is not possible without changes to the nghttp2 C++ bindings source. It needs some extension to the `session` class. I proof-of-concepted it for myself and put a diff to the lib here: https://gist.github.com/dke/99e92ae4b75e0d0520f8cd3a3118cb91

What you can see is that I add methods to make the `SSL*` thing available via a call to the `socket_.native_handle()` method.

Thus the `asio-cl2.cc` (or similar) programs using the C++ bindings can use a new `session.tls_native_handle()` method in the `on_connect` callback access the `SSL*` pointer and e.g. obtain SSL information like `SSL_get_current_cipher(ssl)` etc.

tatsuhiro-t:
Closed since asio library has been removed.

PR #1421, opened 2019-11-25
Expose TLS handle

dke:
Cf https://github.com/nghttp2/nghttp2/issues/1406

I created a merge request to add a method to expose the TLS handle. Usecase is to let the application using the nghttp2 library (or rather, its C++ asio bindings) print some TLS info (protocol name, cipher, tmp key info).


tatsuhiro-t:

PR #1426, opened 2019-12-12
h2load: Allow client to connect to a different target name or IP

LPardue:
This change allows people to use h2load with valid SNI and host/authority but point the client at a designated IP or hostname e.g

```
h2load https://example.com/
h2load --connect-to 127.0.01 https://example.com
h2load --connect-to example.home.arpa https://example.com
```


tatsuhiro-t:
Thank you for PR.  Cherry picked into master via df575f968fa3d82fb40c2637e7d3e245def43766

PR #1435, opened 2020-01-24
Enable session_create_idle_stream test, fix errors

geoffhill:
Add the currently-unused `test_nghttp2_session_create_idle_stream()`
function to the test suite definition.

Modify the test in two places to make it pass:

  * Use stream ID=10 as the priority stream ID to test automatic creation
    of streams for priority specs. The code below checks against stream
    ID=10 so I assume this was a typo in the test.

  * Set the `last_sent_stream_id` instead of the `next_stream_id` to test
    that idle streams cannot be created with smaller numbers than the
    most-recently-seen stream ID. Looking at the validation path in
    `session_detect_idle_stream()`, I think this was another test typo.


tatsuhiro-t:

PR #1444, opened 2020-02-20
Fix receiving stream data stall

tatsuhiro-t:
Previously, if automatic window update is enabled (which is default),
after window size is set to 0 by
nghttp2_session_set_local_window_size, once the receiving window is
exhausted, even after window size is increased by
nghttp2_session_set_local_window_size, no more data cannot be
received.  This is because nghttp2_session_set_local_window_size does
not submit WINDOW_UPDATE.  It is only triggered when new data arrives
but since window is filled up, no more data cannot be received, thus
dead lock happens.

This commit fixes this issue.  nghttp2_session_set_local_window_size
submits WINDOW_UPDATE if necessary.

https://github.com/curl/curl/issues/4939

### DIFF
--- a/lib/nghttp2_session.c
+++ b/lib/nghttp2_session.c
@@ -2494,14 +2494,6 @@ static int session_update_stream_consumed_size(nghttp2_session *session,
 static int session_update_connection_consumed_size(nghttp2_session *session,
                                                    size_t delta_size);
 
-static int session_update_recv_connection_window_size(nghttp2_session *session,
-                                                      size_t delta_size);
-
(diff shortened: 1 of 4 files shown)


dheerajsangamkar:
@tatsuhiro-t , I am using libnghttp2-14 (1.36.0-2) that comes with debian10.
I would like to use this fix. 

Do you think the patch for this fix can be safely backported to nghttp2 version 1.36.0?

tatsuhiro-t:
@dheerajsangamkar If the patch can be applied cleanly, it should work with 1.36.0, but I suggest to use the latest stable version + this fix because it contains the other fixes.

PR #1455, opened 2020-03-31
Fix get_x509_serial for long serial numbers

xjtian:
- Under OpenSSL >= 1.1, `get_x509_serial` would return an empty `StringRef` if the client certificate serial number was greater than 8 bytes long
- Under OpenSSL < 1.1, `get_x509_serial` would return a `StringRef` with the last `(20 - n)` bytes set to nondeterministic garbage and the first `n` bytes set to the correct serial number, where `n` is the length of the serial number.
- This commit unifies the code paths in `get_x509_serial` for all OpenSSL versions and fixes the function to return a `StringRef` of equal length as the serial number.

### DIFF
--- a/src/shrpx_tls.cc
+++ b/src/shrpx_tls.cc
@@ -2030,17 +2030,6 @@ StringRef get_x509_issuer_name(BlockAllocator &balloc, X509 *x) {
 #endif /* !WORDS_BIGENDIAN */
 
 StringRef get_x509_serial(BlockAllocator &balloc, X509 *x) {
-#if OPENSSL_1_1_API && !defined(OPENSSL_IS_BORINGSSL)
-  auto sn = X509_get0_serialNumber(x);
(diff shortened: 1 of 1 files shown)


xjtian:
Note that there's a new compilation warning on this patch:

```
shrpx_tls.cc: In function 'nghttp2::StringRef shrpx::tls::get_x509_serial(nghttp2::BlockAllocator&, X509*)':
shrpx_tls.cc:2044:56: warning: narrowing conversion of 'n' from 'int' to 'size_t {aka long unsigned int}' inside { } [-Wnarrowing]
   return util::format_hex(balloc, StringRef{b.data(), n});
```

I'm not entirely sure how to resolve this one as the return type from `BN_bn2bin` is `int`, not `size_t`.

tatsuhiro-t:
Thank you.  Merged now.

PR #1456, opened 2020-03-31
Add GOAWAY handling on asio client

hap2712:
Upon receiving GOAWAY frames the asio client does not do any signalling towards the user, leaving a stale session object. This has two downsides:
- If there were active streams, session is not stopped but any new messages should not be sent, as they will be rejected by the other endpoint.
- If there were no active streams, session is shutdown silently, leaving the object stale and unusable. The next request will try to create a new stream and will fail with a generic NGHTTP2_INTERNAL_ERROR, which is confusing.

This change attempts to fix that situation by properly calling the error callback with a clear intentional error, signalling that a GOAWAY was received.

### DIFF
--- a/src/asio_client_session_impl.cc
+++ b/src/asio_client_session_impl.cc
@@ -329,6 +329,15 @@ int on_frame_recv_callback(nghttp2_session *session, const nghttp2_frame *frame,
 
     break;
   }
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:

Issue #1389, opened 2019-09-05
nghttp2_hd_inflate_hd_nv() refactoring


yzhao1012:
Hi All,

We'd see how people like the idea of refactoring nghttp2_hd_inflate_hd_nv() such that:
1. There is a API that just decode compressed header fields. And also perform huffman decoding. **But do not perform dynamic table look up.** Static table is fine.
2. Another API then takes the results to perform dynamic table look up.

For #1, the default behavior is to terminate early, if the decoded index is not within the bound of the dynamic table. **But we do need to add the option to continue process.**

For #2, the behavior should be changed to also tolerate certain error, for example, if the dynamic table space is exhausted, the function should have the option to proceed to process other fields. With an option in the input argument, most likely a field inside nghttp2_inflater, to indicate such directive.

The motivation is that, we use nghttp2 for parsing http2 traffic sniffed from network. Being a sniffer, data loss is common, so the dynamic table is no longer accurate. We want to preserve the header indexes for further processing (which I cannot disclose).

ATM, we wrote our own header field decoder, using nghttp2_hd_decode_length(). That is gradually becoming problematic. As we are looking to use huffman decoding code as well, which is not straightforward for integration because its use of internal data structure nghttp2_bufs, which we want to avoid.

I took a look at the code. The changes needed seem to be:
* Augment nghttp2_nv to record index.
* Move out the part the manipulate dynamic table out of nghttp2_hd_inflate_hd_nv() into a new API for #2 above.

I think this refactoring should improve the structure of nghttp2 code. That makes it a win-win situation.

WDYT? Comments on the technical feasibility is also welcome.

AndersBroman:
[nghttp2_wireshark_patch.txt](https://github.com/nghttp2/nghttp2/files/4621788/nghttp2_wireshark_patch.txt)

Hi,
I'm doing the same for Wireshark but I modified/ will modify nghttp2_hd_inflate_hd_nv() to take a
dissection argument 1/0 (TRUE/FALSE) and act accordingly. I have working code. Would this be accepted if offered as a pull request?

Lekensteyn:
[...]

PR #1467, opened 2020-05-13
Create a dissection mode for the deflate function.

AndersBroman:
This is to be used by protocol analysers which wants to show as much as
possible of the header even if packets are missing. In this case a dummy
header is inserted ":Failed deflate", "Index not seen before".

Related https://github.com/nghttp2/nghttp2/issues/1389

### DIFF
--- a/lib/nghttp2_hd.c
+++ b/lib/nghttp2_hd.c
@@ -109,6 +109,10 @@ static const nghttp2_hd_static_entry static_table[] = {
     MAKE_STATIC_ENT("www-authenticate", "", 60, 779865858u),
 };
 
+static const nghttp2_hd_static_entry dissect_index_table[] = {
+    MAKE_STATIC_ENT(":Failed deflate", "Index not seen before", "", -1, NGHTTP2_NV_FLAG_NONE),
+};
+
 static int memeq(const void *s1, const void *s2, size_t n) {
   return memcmp(s1, s2, n) == 0;
 }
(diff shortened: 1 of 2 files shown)


AndersBroman:
I do not know why the travis build fails so I have no way of fixing it.

tatsuhiro-t:
If I understand it correctly, HPACK requires all data to be fed to produce correct output.  If a part of data is missing, its behavior is undefined.  In particular, the table index is not absolute and can be reused.  We have no way to know how many times it is reused.
I think for the purpose of wireshard, it would be nicer to ignore HPACK decompression once HPACK decoder returns error.

AndersBroman:
Hi,
The patch seems to work fine on the traces I have. If the index is reused would that mather from
wiresharks perspective?  The resulting entry would still be Failed deflate. I don't think this is so easy to fix in Wiresharks code as it "saves" the result of the inflate. Also currently I think it refuses
to disssect headers in other streams for the same connection after the inflate failure.

I think that as Wireshark just want to show as much of the header as possible to the user not realy
beeing dependent on the result per se. The aproach this patch takes is acceptable.

lib/nghttp2_hd.c
```diff
@@ -1654,8 +1674,9 @@ static ssize_t hd_inflate_read_len(nghttp2_hd_inflater *inflater, int *rfin,
   }
 
   if (out > maxlen) {
-    DEBUGF("inflatehd: integer exceeded the maximum value %zu\n", maxlen);
-    return NGHTTP2_ERR_HEADER_COMP;
+    DEBUGF("inflatehd: integer %zu exceeded the maximum value %zu\n", out, maxlen);
+    inflater->left = out;
+    return NGHTTP2_ERR_HEADER_COMP_MAX_LEN_EX;
```
Lekensteyn, line 1679: Maybe the tests failed because they expected `inflater->left` to be 0 (or whatever initial state it was) and a return value of `NGTTP2_ERR_HEADER_COMP`?

PR #1468, opened 2020-05-18
fix ubsan errors when compiling with clang-10

asraa:
Running main testsuite with UBSan found some instances of applying zero offsets to null pointers when testing for null frames with zero payloadlen.

runtime error: applying zero offset to null pointer

Signed-off-by: Asra Ali <asraa@google.com>


asraa:
ping @tatsuhiro-t

tatsuhiro-t:
Thank you for PR.  Patch was merged with some modification.
- 78a56cf66f3bbe29f70b19eeaff0dbc63908c677
- cf094bd56bd3fbe0037fc72a2ebcee3cb7724b90

PR #1488, opened 2020-06-17
doc: clarify flow control behaviour for nghttp2_session_send()

tomaskrizek:
The previous wording implied that if send_callback doesn't return
NGHTTP2_ERR_WOULDBLOCK, this function would always send all
submitted requests / data. This is not the case, since flow control can
cause some data to remain queued until WINDOW_UPDATE is received.


tatsuhiro-t:
Merged now.  Thank you for PR.

PR #1573, opened 2021-03-21
CRUD operation support

wallyatgithub:
This request adds CRUD operation to h2load:

support varible in URI and data file, and replace it with a value in given range in config when send the request

monitor the response for a header matching given name from config, and find the valid uri from it

send read/update/delete to the uri found in step 2, with method for each operation type configurable

add stream timeout support, send RST_STREAM if no response comes in given interval (configurable)

print RPS and success rate every second

fix bug around --rps_req_inflight


wallyatgithub:
Hi @tatsuhiro-t ,

Can you review this pull request?


Thanks a lot

PR #1576, opened 2021-04-03
Fix segfault for for session shutdown

OlexiyKhokhlov:
I would like to call shutdown() in the session on_error handler.
Sometimes it provides a crash because session_impl::session_  can be nullptr
I guess it will be more comfortable to check it inside the shutdown() call


tatsuhiro-t:

PR #1595, opened 2021-06-23
enable asio client to attach to an existing stream

mosaics14:
The purpose of the change to asio client is to have the ability to attach to an existing stream. One use case is that caller can establish an tunnel through a proxy, then start http/2 over the tunnel.


tatsuhiro-t:

Issue #1600, opened 2021-07-31
Is it possible to pause one stream on receive data from server


nylon009:
The device has very limited memory, so we can not download all the response together,  we need to pause some stream and resume it later. 

For exampele, we play audio on device, we can not download all the audio, need to do it in a streaming way.

We set a ringbuffer to cache the data, if the ringbuffer is full,  we need to pause the download of the audio stream, but we cannot block nghttp2_on_data_chunk_recv_callback function to avoid affect other streams. 
if the ringbuff is not full, we notify the nghttp2 session to resume the stream.

JackyYin:
Hi, how about check stream ID before pause it?
something like this:
```
 if (is_ringbuffer_full(stream_id)) {
    return  nghttp2_error.NGHTTP2_ERR_PAUSE;
}
```
according to the document:
> If the application uses nghttp2_session_mem_recv(), it can return nghttp2_error.NGHTTP2_ERR_PAUSE to make nghttp2_session_mem_recv() return without processing further input bytes. The memory by pointed by the data is retained until nghttp2_session_mem_recv() or nghttp2_session_recv() is called.

JackyYin:
Hi, I have modified examples/libevent-client.c to test this and I think it works.
I am able to read RING_BUFFER_LEN of data into my buffer and process it.
sleep(2) is just for simulate time-consuming processing.

readcb:
```
static void readcb(struct bufferevent *bev, void *ptr) {
  http2_session_data *session_data = (http2_session_data *)ptr;
  ssize_t readlen;
  struct evbuffer *input = bufferevent_get_input(bev);
  size_t datalen = evbuffer_get_length(input);
  unsigned char *data = evbuffer_pullup(input, -1);

  session_data->accureadlen = 0;
  while (session_data->accureadlen < datalen) {
    readlen = nghttp2_session_mem_recv(session_data->session, data, MIN(datalen, RING_BUFFER_LEN));
    if (readlen < 0) {
      warnx("Fatal error: %s", nghttp2_strerror((int)readlen));
      delete_http2_session_data(session_data);
      return;
    }
    session_data->accureadlen += readlen;
    sleep(2);
  }
  if (evbuffer_drain(input, (size_t)datalen) != 0) {
    warnx("Fatal error: evbuffer_drain failed");
    delete_http2_session_data(session_data);
    return;
  }
  if (session_send(session_data) != 0) {
    delete_http2_session_data(session_data);
    return;
  }
}
```

[...]

PR #1603, opened 2021-08-09
update doc for nghttp2_session_mem_recv

JackyYin:
I think nghttp2_session_mem_recv is able to receive only inlen bytes of data instead of all.
I am trying to make the document clearer.
Also expect to update this page too: [nghttp2_session_mem_recv](https://nghttp2.org/documentation/nghttp2_session_mem_recv.html#c.nghttp2_session_mem_recv)

reference issue:  #1600


tatsuhiro-t:

PR #1613, opened 2021-08-31
Stricter checks for pseudo-headers :method and :path

### LINKED ISSUE #1611: Implement stricter checks for allowed characters in pseudo-header values
The checks for the allowed characters in pseudo-headers like ":method" and ":path" should be improved.

[":method" must be a token](https://datatracker.ietf.org/doc/html/rfc7231#section-4). The definition for "token" is [here](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6). Note that e.g. spaces and tabs are not allowed, and the value must not be empty.  (Update: the "not empty" check is already implemented.)

[The rules for ":path" are more complex](https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.3), but here also spaces and tabs are not allowed, and the value must not be empty for http/https. (Update: the "not empty" check is already implemented.)

These checks are important to prevent HTTP/2 downgrading attacks. Please read the research paper [HTTP/2: The Sequel is Always Worse](https://portswigger.net/research/http2) for more information.

mkauf:
Check the allowed characters for ":method" (see RFC 7230, section 3.2.6) and
":path". For ":path", the space and tab characters are now forbidden, but
other special characters are still allowed for compatibility reasons.

Update genvchartbl.py so that it generates the same table as in the code.

Fixes #1611


tatsuhiro-t:
Thank for PR.  Merged now.

Issue #1639, opened 2021-11-13
h2load cannot set max_frame_size


robaho:
If the default frame size of 16k is too small for high throughput transfers.

There is no way to set the receiver frame size using h2load.

robaho:
added https://github.com/nghttp2/nghttp2/pull/1640 to enable this

robaho:
I think the documentation [here](https://nghttp2.org/documentation/h2load-howto.html#flow-control) needs to be updated as well.

tatsuhiro-t:
Why?  The size of frame is irrelevant to flow control.

robaho:
Because it is a guide, and the section is not the help, but is offered with "and it may affect benchmarking results".

The frame size will affect the benchmarking results for large transfers by 5x-10x and I think it should be noted (and probably also discussed why - e.g. reduces contention in concurrency by reducing the number of frames needed to be processed).

tatsuhiro-t:
Could you give me the evidence that the large frame size offers 10x performance increment?  with the detailed client and server setup and configurations.

robaho:
The full discussion can be found [here](https://groups.google.com/d/msgid/golang-nuts/89926c2f-ec73-43ad-be49-a8bc76a18345n%40googlegroups.com).

In general, as compared to Http1, the framing of Http2 is overhead. By increasing the frame size you lower the overhead per byte transmitted for large transfers.

It may be difficult to detect the overhead unless you are working on a very fast network. In lieu of that, running the client and server on the same machine using local networking will get you close (although with modern networks the CPU overhead of running both may be worse).

Using an iMac with i7 quad-core processor, I obtained the following results:

> Using a Go server with default frame size (16k):
> 
> Go client:  900 MB/s
> Java client: 1300 MB/s
> curl: 1500 MB/s
> 
> Using a Java server with default frame size (16k):
> 
> Go client: 670 MB/s
> Java client: 720 MB/s
> curl: 800 M/s
> 
> Using Go server using 256k client max frame size:
> 
> Go client: 2350 MB/s
> Java client: 2800 MB/s
> h2load: 4300 MB/s
> 
> Using Java server using 256k client max frame size:
> 
> Go client: 2900 MB/s
> Java client: 2800 MB/s
> h2load: 3750 MB/s

robaho:
So, in the above case, it is not quite 10x, closer to 5x from worst to best using Go. This is representative because the overhead of frame handling in the Go implementation is higher than in h2load.

[...]

PR #1640, opened 2021-11-14
allow setting max frame size for h2load

robaho:
The http2 spec allows for clients to set the max frame size. This is essential for high-speed high-throughput downloads.

Fixes issue #1639.


tatsuhiro-t:
Thank you for PR.  Cherry-picked as f92f81c05a656e6f84ba281403ca8a0207a7e6f6 with some amends.

Issue #1661, opened 2021-12-25
[asio_server.cc] data race on server.stop() and exception thrown sometimes


francois-random:
Hello,
there is a data race when using acceptor.close() (acceptor is not thread safe) while still listening from another thread (server.close() is by design used from a thread not in the threading pool).
It can cause sometimes "terminate called after throwing an instance of 'boost::wrapexcept<boost::system::system_error>'
  what():  set_option: Bad file descriptor" when reaching  the line "new_connection->socket().set_option(tcp::no_delay(true));" .
Because of the data race, it's possible to be in the handler of async_accept with both error_code that contains no error AND new_connection->socket().is_open() which is FALSE.
(set_option on a socket which is not opened will throw an exception)

see this simple example to reproduce:

```
class dummy_listen
{
public:
    dummy_listen ( boost::asio::io_context& context, unsigned int counter )
        : _acceptor ( context ), _counter ( counter )
    {
    }
    bool start_listen ( unsigned short port )
    {
        boost::asio::ip::tcp::endpoint endpoint ( boost::asio::ip::address_v4::any(), port );
        _acceptor.open ( endpoint.protocol() );
        _acceptor.set_option ( boost::asio::socket_base::reuse_address ( true ) );
        boost::system::error_code ec;
        _acceptor.bind ( endpoint, ec );
        if ( ec ) {
            // An error occurred.
            _acceptor.close ( ec );
            return false;
        }
        _acceptor.listen ( 128, ec );
        if ( ec ) {
            // An error occurred.
            _acceptor.close ( ec );
            return false;
        }
        return true;
    }
    bool safe_stop()
    {
        std::promise<bool> promise;
        boost::asio::dispatch ( _acceptor.get_executor(),
                                [this, &promise]() {
                                    boost::system::error_code ec;
                                    _acceptor.close ( ec );
                                    promise.set_value ( !ec );
                                } );
        return promise.get_future().get();
    }
    bool bad_stop()
    {
        boost::system::error_code ec;
        _acceptor.close ( ec );
        return !ec;
    }
    void start_accept()
    {
        auto socket = std::make_shared<boost::asio::ip::tcp::socket> ( _acceptor.get_executor() );
        _acceptor.async_accept ( *socket, [=] ( const boost::system::error_code& ec ) {
            if ( !ec ) {
                if ( !_acceptor.is_open() ) {
                    std::cerr << "warning for listener " << _counter << " : acceptor is not open!" << std::endl;
                }
[...]

PR #1662, opened 2021-12-25
prevents data race in server.stop() - see bug #1661

francois-random:
prevents data race in server.stop() - see bug #1661


tatsuhiro-t:

PR #1679, opened 2022-03-11
FIX_BEFORE_MERGE: stream can get stuck when SETTINGS_INITIAL_WINDOW_SIZE lowers window size

Natris:
I am hitting this in my unit tests where I lower Initial window size.

Client connects to server. Server operating in manual window update mode sends SETTINGS. Client sends request headers and a bunch of data that does not yet require WINDOW_UPDATE to be sent. Client sends SETTINGS ACK; this applies new settings. This makes window completely full for the client; however server does not check nghttp2_should_send_window_update from within nghttp2_session_update_local_settings. This leads to stream being stuck.

The testcase included emulates what was described above; it is perhaps too complicated.

CAUTION: this fix breaks 
  Test: session_submit_settings_update_local_window_size ...FAILED
    1. .../nghttp2_session_test.c:5800  - 32768 == stream->recv_window_size
    2. .../nghttp2_session_test.c:5804  - 0 == nghttp2_session_get_stream_local_window_size(session, 1)
This needs to be fixed somehow...

### DIFF
--- a/lib/nghttp2_session.c
+++ b/lib/nghttp2_session.c
@@ -4259,8 +4259,7 @@ static int update_local_initial_window_size_func(void *entry, void *ptr) {
     return nghttp2_session_add_rst_stream(arg->session, stream->stream_id,
                                           NGHTTP2_FLOW_CONTROL_ERROR);
   }
-  if (!(arg->session->opt_flags & NGHTTP2_OPTMASK_NO_AUTO_WINDOW_UPDATE) &&
-      stream->window_update_queued == 0 &&
+  if (stream->window_update_queued == 0 &&
       nghttp2_should_send_window_update(stream->local_window_size,
                                         stream->recv_window_size)) {
 
--- a/tests/main.c
+++ b/tests/main.c
@@ -196,6 +196,8 @@ int main(void) {
                    test_nghttp2_submit_settings) ||
       !CU_add_test(pSuite, "session_submit_settings_update_local_window_size",
                    test_nghttp2_submit_settings_update_local_window_size) ||
+      !CU_add_test(pSuite, "test_nghttp2_submit_settings_update_local_window_size_before_initial_settings_ack",
+                   test_nghttp2_submit_settings_update_local_window_size_before_initial_settings_ack) ||
       !CU_add_test(pSuite, "session_submit_settings_multiple_times",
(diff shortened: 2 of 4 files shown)


tatsuhiro-t:
Smells like a corner case bug.  I'll look into it.

tatsuhiro-t:
The correct fix would be something like this:

```patch
diff --git a/lib/nghttp2_session.c b/lib/nghttp2_session.c
index 380a47c1..2cb2ad31 100644
--- a/lib/nghttp2_session.c
+++ b/lib/nghttp2_session.c
@@ -4247,6 +4247,10 @@ session_update_remote_initial_window_size(nghttp2_session *session,
                           update_remote_initial_window_size_func, &arg);
 }
 
+static int session_update_stream_consumed_size(nghttp2_session *session,
+                                               nghttp2_stream *stream,
+                                               size_t delta_size);
+
 static int update_local_initial_window_size_func(void *entry, void *ptr) {
   int rv;
   nghttp2_update_window_size_arg *arg;
@@ -4259,9 +4263,16 @@ static int update_local_initial_window_size_func(void *entry, void *ptr) {
     return nghttp2_session_add_rst_stream(arg->session, stream->stream_id,
                                           NGHTTP2_FLOW_CONTROL_ERROR);
   }
-  if (!(arg->session->opt_flags & NGHTTP2_OPTMASK_NO_AUTO_WINDOW_UPDATE) &&
-      stream->window_update_queued == 0 &&
-      nghttp2_should_send_window_update(stream->local_window_size,
+
+  if (stream->window_update_queued) {
+    return 0;
+  }
+
+  if (arg->session->opt_flags & NGHTTP2_OPTMASK_NO_AUTO_WINDOW_UPDATE) {
+    return session_update_stream_consumed_size(arg->session, stream, 0);
+  }
+
+  if (nghttp2_should_send_window_update(stream->local_window_size,
                                         stream->recv_window_size)) {
 
     rv = nghttp2_session_add_window_update(arg->session, NGHTTP2_FLAG_NONE,
```

Natris:
yes; I must admit that since I do not use auto window update mode, I did not even attempt to dive deep into that logic :)

This thing has been plaguing our tests for years, as it is very rare and any attempt to increase level of logging usually makes it disappear as it is timing sensitive :(

PR #1690, opened 2022-03-29
*NOT READY FOR MERGING* Add grpc server outline

pyssling:
Only handles unary grpcs. Only success implemented this far.

This is an outline for discussion:
- Does the project want this?
- If so - how far should we take this.

Still to do:
- Different statuses than success
- Move parts to impl file
- Code generation from protoc
- Lots more...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -242,6 +242,10 @@ set(HAVE_JEMALLOC   ${JEMALLOC_FOUND})
 
 if(ENABLE_ASIO_LIB)
   find_package(Boost 1.54.0 REQUIRED system thread)
+
+  if(ENABLE_GRPC)
+      find_package(Protobuf REQUIRED)
+  endif()
 endif()
 
 # libbpf (for bpf)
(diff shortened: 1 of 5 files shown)


pyssling:
@tatsuhiro-t any thoughts on this? If you feel it doesn't fit with the library I can implement this in a separate repository.

pyssling:
Hi, no worries. This is more of a hobby project than anything else.
I understand about the asio lib - if I can be a bit critical - hopefully constructively so and mostly not against nghttp2 really:
- boost/asio is so unstable abi/api wise that shipping a binary library (which asio/nghttp2 is) doesn't make a lot of sense. It would be better to ship this as a separate header only library that tries to maintain some compatibility against multiple versions of boost/asio.
- Some of the abstractions in the library are coupled to old versions of boost/asio - I think remoddeling it would be a good idea - specifically the thread/multiple io_service/io_context handling on the server side is not coherent without how modern boost/asio does things.

Some other questions/reflections:
- Would you be interested in a lite grpc implementation in C? Obviously not using the protobuf API, but just handling framing/status/application header? This would be quite easy to do.
- Are you ok with me copying out the asio lib and doing some cleanup as a separate library? We could maybe remerge later if you wish, or it could just remain separate. Obviously license would remain the same.

tatsuhiro-t:
Yeah, basically, C++ lib is quite hard to maintain its ABI, the only practical form is header only.  I have no time to follow recent activities of boost stuff.

> Would you be interested in a lite grpc implementation in C? 

I'm not that interested in grpc now.  In my opinion, grpc is actually HTTP/2+something, not pure HTTP/2, so better to use their own libraries.

> Are you ok with me copying out the asio lib and doing some cleanup as a separate library?

It is fine as long as the way our license permits.  It is MIT, very easy to pass those restrictions.
If your library goes successful, I would remove asio lib from this repository.

tatsuhiro-t:

PR #1771, opened 2022-08-04
fuzz: add extended fuzzer using FuzzedDataProvider

DavidKorczynski:
Similar in nature to fuzz_target.cc but adds the difference of a random
iv constructed by way of fuzzer data. This fuzzer achieves higher code
coverage than fuzz_target.cc but I think we should create a new rather
than extend the existing to keep corpus of the existing.


tatsuhiro-t:
Thank you for PR.  KGTM

PR #1775, opened 2022-08-09
Replace the use of strtoul and strtol with parse_uint

### LINKED ISSUE #1774: Abnormal Parameters in the nghttpd Command
hello，I use the nghttpd command line to start an HTTP/2 server. However, I find that if an abnormal parameter, for example, nghttpd --no-tls 8093 -v -n -1, appears in the command line, a coredump occurs. This parameter should be verified during parameter reading.
image
image

tatsuhiro-t:
Replace the use of strtoul and strtol with parse_uint to fix the
handling of negative integer.

Fixes #1774


patlecat:
Why did you not use the fmt lib's conversion?

PR #1792, opened 2022-09-06
Option no rfc9113 ltws validation

tatsuhiro-t:
Escape hatch to allow leading and trailing white spaces in field values.
cc @bagder @icing


bagder:
I will probably enable this by default in curl, seeing there seem to be so many violations...

icing:
Thank you, @tatsuhiro-t. I'll make this configurable, but will have for backward compatibility reasons default to the previous, rfc7540 checks.

PR #1802, opened 2022-09-23
Update dependency versions for Dockerfile.android

PufferBlue:
1.40.0 is the last version that can make a successful build with this Dockerfile, build failed start from 1.41.0
Updated some dependency versions
Fixed a typo


tatsuhiro-t:

PR #1824, opened 2022-10-31
cmake: make the .pc file relocatable

lazka:
Instead of writing absolute file paths for exec_prefix, libdir and includedir make them all relative to prefix. This makes the .pc file match the autotools generated one and allows changing all paths via prefix.

Before:

prefix=C:/test
exec_prefix=C:/test
libdir=C:/test/lib
includedir=C:/test/include

After:

prefix=C:/test
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include


lazka:
(someone mentioned CMAKE_INSTALL_LIBDIR can be absolute. I'll update this PR to handle that case)

lazka:
> (someone mentioned CMAKE_INSTALL_LIBDIR can be absolute. I'll update this PR to handle that case)

done now, better late then never :)

PR #1961, opened 2023-10-10
Rework session management

tatsuhiro-t:


tatsuhiro-t:
yes

hippohobbily:
Isn't initializing rl->tstamp = 0 going to cause the first time nghttp2_ratelim_update() is called to pass in a tstamp that's very big?
Then the delta would be == the tstamp and I don't think it's going to work from then on.
Should it be initialized to nghttp2_time_now_sec()?

PR #1967, opened 2023-10-14
cmake: speed up warning option detection

vszakats:
This patch makes CMake configuration faster by dropping feature checks
for each individual "picky" C/C++ compiler warning options and instead
enable them based on compiler version. It also allows to add options using
the old, feature-check method.

I've implemented this for libssh2 [1], curl [2] and ngtcp2 [3] earlier.
This patch is based on those, by adapting it to the warning options used
by nghttp2.

It brings CMake configuration time down from 22 to 15 seconds on my local
machine.

Also add missing `WARNCXXFLAGS` to the build config summary.

[1] libssh2
https://github.com/libssh2/libssh2/commit/ec0feae7920d695ce234a5aba13014bf29824c09
https://github.com/libssh2/libssh2/pull/952

[2] curl
https://github.com/curl/curl/commit/9c543de0ecf14880c3324d3d24591fb695dd1917
https://github.com/curl/curl/pull/10973

[3] ngtcp2
https://github.com/ngtcp2/ngtcp2/pull/964

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -339,74 +339,12 @@ if(CMAKE_C_COMPILER_ID MATCHES "MSVC")
   endif()
 else()
   if(ENABLE_WERROR)
-    extract_valid_c_flags(WARNCFLAGS    -Werror)
-    extract_valid_c_flags(WARNCXXFLAGS  -Werror)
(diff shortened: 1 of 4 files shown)


tatsuhiro-t:

PR #1996, opened 2023-11-07
Have less stuff in config.h

hrxi:
Speeds up the build minimally, doesn't check for stuff that's always there.


neheb (approved):

neheb:
Ah right. Needs automake changes too.

neheb (approved):

tatsuhiro-t:

PR #2011, opened 2023-12-14
cmake: bring back ENABLE_STATIC_CRT

bwncp:
While `ENABLE_STATIC_CRT` is still present in CMakeOptions.txt, it has been removed by 81c199dd. I think this was by accident.

Add it again, otherwise our Windows builds will fail.


tatsuhiro-t:

vszakats:
Sorry about deleting this, it wasn't intended.

PR #2028, opened 2024-01-03
bump openssl min required version to 3.0

chenrui333:
openssl 1.1 is already eol, bump min required version to 3.0 (which is LTS)

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,7 @@ if(NOT ENABLE_LIB_ONLY)
   find_package(ZLIB 1.2.3)
 endif()
 
-find_package(OpenSSL 1.1.1)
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
I leave minimum version 1.1.1 because it works with boringssl and libressl.  As far as I know, the only working version of openssl is 1.1.1.  OpenSSL 3.x have critical flaws such as degraded performance, too much locking, unintuitive text based configuration runtime etc.  I suggest to migrate to another TLS stack, like aws-lc, libressl, boringssl.  wolfSSL support will be added soon.

anthonyalayo:
I agree with @tatsuhiro-t , OpenSSL is pinned for my projects at 1.1.1 for the same reason.

kickinchicken (approved):

PR #2030, opened 2024-01-08
Update minimum CMake version to 3.5

anthonyalayo:
On the latest version of CMake, a warning is emitted if it detects a project using min version < 3.5:
```
CMake Deprecation Warning at build/_deps/nghttp2-src/CMakeLists.txt:25 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

CMake 3.5 was released in March 2016, so having it upgraded seems reasonable:
https://cmake.org/pipermail/cmake/2016-March/062947.html


anthonyalayo:
cc @tatsuhiro-t

anthonyalayo:
CI checks passed, let me know what you think @tatsuhiro-t

tatsuhiro-t:

anthonyalayo:
@tatsuhiro-t when is your next version release of nghttp2 planned?

PR #2064, opened 2024-02-12
Munit

tatsuhiro-t:

PR #2092, opened 2024-03-02
Respect BUILD_STATIC_LIBS and add option for tests

### LINKED ISSUE #2088: Static library always built with identical lib name as shared causing linking errors
As you have removed the ``if(HAVE_CUNIT OR ENABLE_STATIC_LIB)`` condition in version 1.60.0, it's always building the static library, even though I explicitly have been disabling it with ``ENABLE_STATIC_LIB=OFF``, and since 1.60.0 with ``BUILD_STATIC_LIBS=OFF``. ``STATIC_LIB_SUFFIX`` is never set, but is used here: https://github.com/nghttp2/nghttp2/blob/c5e087b1038b9b0a972ca07748a525b8f58139d9/lib/CMakeLists.txt#L74
But without ``STATIC_LIB_SUFFIX`` being set, the static library is using identical lib name as the shared lib causing linking errors, as you can see here: https://github.com/strawberrymusicplayer/strawberry-msvc-dependencies/actions/runs/8115697967/job/22184102638?pr=370
This means the user has to configure CMake with ``STATIC_LIB_SUFFIX`` to avoid the conflict when building a shared library.
The ``if(BUILD_STATIC_LIBS)`` line should be moved so it does not only apply to the install, but building too.
And ``STATIC_LIB_SUFFIX`` should probably be CMake option defaulting to "static" so it does create conflicts when building both shared and static.

jonaski:
Respect BUILD_STATIC_LIBS and add option for tests.

This also fixes the library conflict with MSVC by setting STATIC_LIB_SUFFIX so it's using "nghttp2_static" when both static and shared library is built, since the lib name currently conflicts with the shared when STATIC_LIB_SUFFIX was not set.

Fixes #2088


tatsuhiro-t:
Thank you for PR.  The failure of CIFuzz needs fixing oss-fuzz build script. I'll do that later.

For those mentioned recent cmake issues, @jmckenna @anthonyalayo @vszakats, this PR changes the default static library name from libnghttp2.a to libnghttp2_static.a.  Are you fine with this?

lib/CMakeLists.txt
```diff
@@ -66,23 +66,22 @@ if(BUILD_SHARED_LIBS)
 endif()
 
 # Static library (for unittests because of symbol visibility)
-add_library(${STATIC_LIB} STATIC ${NGHTTP2_SOURCES})
+if(BUILD_STATIC_LIBS)
```
anthonyalayo, line 69: reasonable to me, this got removed in:
https://github.com/nghttp2/nghttp2/pull/2064/files#diff-4803677bd280f4dded52c428e3c8ac6c618508dcabcbe880b68be85897c2ad41L68

CMakeLists.txt
```diff
@@ -432,9 +432,11 @@ add_subdirectory(lib)
 add_subdirectory(third-party)
 add_subdirectory(src)
 add_subdirectory(examples)
-add_subdirectory(tests)
-#add_subdirectory(tests/testdata)
-add_subdirectory(integration-tests)
+if(BUILD_STATIC_LIBS)
```
anthonyalayo, line 435: Ideally the control flow is a little better than this, since building static libs shouldn't imply that one wants to also build testing.

Can we make this either:
```
if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME AND BUILD_STATIC_LIBS)
  add_subdirectory(tests)
  #add_subdirectory(tests/testdata)
  add_subdirectory(integration-tests)
endif()
```

Or create a new CMake option, `BUILD_TESTING`, and do this:
```
if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME AND BUILD_TESTING)
  add_subdirectory(tests)
  #add_subdirectory(tests/testdata)
  add_subdirectory(integration-tests)
endif()
```

where `BUILD_TESTING` is defined as:
```
cmake_dependent_option(BUILD_TESTING "Enable tests" ON "BUILD_STATIC_LIBS" OFF)
```

I'm fine with either way. cc @tatsuhiro-t

CMakeOptions.txt
```diff
@@ -12,7 +12,7 @@ option(ENABLE_EXAMPLES  "Build examples"
 option(ENABLE_FAILMALLOC "Build failmalloc test program" ON)
 option(ENABLE_LIB_ONLY  "Build libnghttp2 only.  This is a short hand for -DENABLE_APP=0 -DENABLE_EXAMPLES=0 -DENABLE_HPACK_TOOLS=0")
 option(BUILD_SHARED_LIBS "Build libnghttp2 as a shared library" ON)
-option(BUILD_STATIC_LIBS "Build libnghttp2 in static mode also" OFF)
+option(BUILD_STATIC_LIBS "Build libnghttp2 in static mode also" ON)
```
anthonyalayo, line 15: I'm not following exactly why this needs to be on by default. I'm fine either way, but if the reasoning is that testing needs it, all one needs to do is pass `-DBUILD_STATIC_LIBS=ON`. What's the motivation?

anthonyalayo (commented):
Static library rename is reasonable to me, but I left some comments on other parts that could be improved

anthonyalayo (approved):

anthonyalayo:
LGTM. @tatsuhiro-t ?

lib/CMakeLists.txt
```diff
@@ -66,23 +66,22 @@ if(BUILD_SHARED_LIBS)
...
+  add_library(${STATIC_LIB} STATIC ${NGHTTP2_SOURCES})
 
-set_target_properties(${STATIC_LIB} PROPERTIES
-  COMPILE_FLAGS "${WARNCFLAGS}"
-  VERSION ${LT_VERSION} SOVERSION ${LT_SOVERSION}
-  ARCHIVE_OUTPUT_NAME nghttp2${STATIC_LIB_SUFFIX}
```
vszakats, line 74: Please don't rename static libs. This has been a source of a list of issue in the past, and changing will just restart this all over again.

vszakats (changes requested):
Please don't rename the static lib. This has been a source of a list of issues in the past, and changing it will just restart this all over again.

Honestly I also don't understand the point in silently renaming the static/shared enabler options... Now they differ from sister projects nghttp3 and ngtcp2. This change was unrelated to the commit it was mixed into. And now continuing with another rename in a fixup patch!

Also notice that `STATIC_LIB_SUFFIX` is there to give users control over the name of static lib. This PR now removes this option. When building both shared and static, this value can be set to something to avoid collision (required for MSVC only). Alternatively it could be set automatically in such cases to avoid the collision by default. This is implemented in e.g. in curl like so: https://github.com/curl/curl/blob/ae7ad31be209b8c0168c213a4dbc4f869f177823/lib/CMakeLists.txt#L95-L101

anthonyalayo:
> Please don't rename the static lib. This has been a source of a list of issues in the past, and changing it will just restart this all over again.

Out of curiosity, what was the list of issues in the past? Most projects I know do have separate names for them.

vszakats:
Summarized here: https://github.com/curl/curl/pull/7515#issuecomment-890320856

Renaming and deleting variables or changing default build behaviour always stir up downstream builds. In this particular case this round already happened in the past, due to wrong assesment of the issue and an inaccurate fix as a result.

There is no universal lib naming convention in the MSVC arena, so hard-coding anything will not work for some/most downstream projects, while breaking all existing ones. I added `STATIC_LIB_SUFFIX` to cover this in a generic way, but it must be set manually to the desired suffix when building both static and shared.

Another option is to rename the implib in this case, but that's similarly arbitary, so we need to keep it open for a manual override.

anthonyalayo:
Interesting. I haven't dabbled in the MSVC world, but I would imagine downstream projects would also be using CMake (I know curl does). If they do, the CMake config files usually takes care of everything such that the naming of libs don't matter. But I can imagine there is a lot of other "integrations" with the build output such that some naming signature is needed. Makes sense.

vszakats (approved):

tatsuhiro-t:
Now static lib is not build by default and therefore testing support is being off.  We need test to run on CI.
I think `-DBUILD_STATIC_LIBS=ON` should be added to:
- https://github.com/nghttp2/nghttp2/blob/ac2e671d80fdc2268850c3cfd920efb1ede408ec/.github/workflows/build.yml#L423
- https://github.com/nghttp2/nghttp2/blob/ac2e671d80fdc2268850c3cfd920efb1ede408ec/.github/workflows/build.yml#L436 ; and
- https://github.com/nghttp2/nghttp2/blob/ac2e671d80fdc2268850c3cfd920efb1ede408ec/.github/workflows/build.yml#L526

tatsuhiro-t:
OK, now fuzzing check passes.

jmckenna:
looks good here

tatsuhiro-t:
Thank you for everyone involved.  Merged now.

PR #2093, opened 2024-03-02
Checkout with submodules

jonaski:
Reduce commands by using `actions/checkout` with "`submodules: recursive`", also does the same for `git clone` commands.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,8 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-    - uses: actions/checkout@v4
+    - name: Checkout
+      uses: actions/checkout@v4
     - name: Restore libbpf cache
       id: cache-libbpf
       uses: actions/cache@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -516,14 +518,14 @@ jobs:
...
+        fetch-depth: 1
+        submodules: recursive
     - uses: microsoft/setup-msbuild@v2
     - name: Configure cmake
-      run: |
-        git submodule update --init --depth 1
-        mkdir build
-        cd build
-        cmake -DCMAKE_TOOLCHAIN_FILE=C:/vcpkg/scripts/buildsystems/vcpkg.cmake -DCMAKE_GENERATOR_PLATFORM=${{ matrix.platform }} -DVCPKG_TARGET_TRIPLET=${{ matrix.arch}}-windows ..
+      run: cmake -S . -B build -DCMAKE_TOOLCHAIN_FILE=C:/vcpkg/scripts/buildsystems/vcpkg.cmake -DCMAKE_GENERATOR_PLATFORM=${{ matrix.platform }} -DVCPKG_TARGET_TRIPLET=${{ matrix.arch}}-windows ..
```
tatsuhiro-t, line 528: If I understand it correctly, `-S .` can be omitted here.
And `..` should be removed.

tatsuhiro-t:

PR #2109, opened 2024-03-20
Automate release process

tatsuhiro-t:


aduh95:
Where can we find the public key to verify the signatures? I wasn't able to find the info in the documentation

PR #2128, opened 2024-03-31
cmake: check `SSL_provide_quic_data` when `ENABLE_HTTP3` is `ON`

jimmy-park:
The `SSL_provide_quic_data` symbol is always checked when OpenSSL is found. The official OpenSSL doesn't have `SSL_provide_quic_data`, so the warning message is printed unnecessarily.


tatsuhiro-t:
Yes, that is unfortunate.  Because openssl 3.x has degraded performance, it would be better to build nghttp2 with boringssl or aws-lc for now.

tatsuhiro-t:

PR #2249, opened 2024-09-23
fix race condition on h1 connection close

TuxInvader:
Running a duration test over http/1.1 with connection close, will usually fail to complete. Eg: 
```
h2load -c50 -t1 -D10 --h1 --connect-to 10.10.0.148:4443 -H "Connection: close"  https://www.foo.com/0k.html
```
This is because there is a race condition between a client being terminated and it spawning a new connection. This patch prevents new clients being launched after entering the `DURATION_OVER` phase.


tatsuhiro-t:

PR #2268, opened 2024-10-25
account for bytes on closing connections

TuxInvader:
If a connection is closing it returns -1 from on_read() and so the bytes in the last response are not counted in the stats.

### DIFF
--- a/src/h2load.cc
+++ b/src/h2load.cc
@@ -1242,11 +1242,13 @@ int Client::connection_made() {
 
 int Client::on_read(const uint8_t *data, size_t len) {
   auto rv = session->on_read(data, len);
+  if (worker->current_phase == Phase::MAIN_DURATION) {
+    if ( len > 0 ) {
+      worker->stats.bytes_total += len;
+    }
+  }
   if (rv != 0) {
     return -1;
-  }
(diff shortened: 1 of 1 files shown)


src/h2load.cc
```diff
@@ -1242,12 +1242,14 @@ int Client::connection_made() {
 
 int Client::on_read(const uint8_t *data, size_t len) {
   auto rv = session->on_read(data, len);
+  if (worker->current_phase == Phase::MAIN_DURATION) {
+    if ( len > 0 ) {
```
tatsuhiro-t, line 1246: I think this is not needed.  The previous code does not have this condition.

TuxInvader: Maybe not, but I wasn't sure if it would be possible for `len` to ever be a negative number. 
So, I was trying to avoid the possibility of bytes being deducted.

tatsuhiro-t: len is of type size_t that is unsigned integer, never become negative.

TuxInvader: Oops. Removed the unnecessary check. Thanks.

tatsuhiro-t:

PR #2271, opened 2024-11-05
Workaround the lack of ssize_t type under Windows

georgthegreat:
As [MSDN says](https://learn.microsoft.com/en-us/windows/win32/winprog/windows-data-types):

```
A signed version of SIZE_T.
This type is declared in BaseTsd.h as follows:
typedef LONG_PTR SSIZE_T;
```

### DIFF
--- a/lib/includes/nghttp2/nghttp2.h
+++ b/lib/includes/nghttp2/nghttp2.h
@@ -55,6 +55,11 @@ extern "C" {
 
 #include <nghttp2/nghttp2ver.h>
 
+#if defined(_MSC_VER)
+#  include <basetsd.h>
+   typedef SSIZE_T ssize_t;
+#endif
+
 #ifdef NGHTTP2_STATICLIB
 #  define NGHTTP2_EXTERN
 #elif defined(WIN32) ||                                                        \


tatsuhiro-t:
The official solution is use of nghttp2_ssize https://github.com/nghttp2/nghttp2/blob/7a96731c6b03810bb5f45b3bd34806b895d7446e/lib/includes/nghttp2/nghttp2.h#L84

georgthegreat:
@tatsuhiro-t it might be the correct solution, but there is 15 occurances of `ssize_t` in `nghttp2.h`, hence the code just does not compile when targeting Windows.

tatsuhiro-t:
Defining NGHTTP2_NO_SSIZE_T eliminates usage of ssize_t and then it compiles fine on platform that lacks ssize_t.

georgthegreat:
I wonder what is the rationale behind this?
If there is a ssize_t on Windows and the library wants to use it, why do we need to customize the build at this point?

georgthegreat:
And no, passing `-DNGHTTP2_NO_SSIZE_T` via CFLAGS does not work for me either.

georgthegreat:
This code breaks the suggested behavior:

```
#ifdef BUILDING_NGHTTP2
#  undef NGHTTP2_NO_SSIZE_T
#endif
```

tatsuhiro-t:
To clarify:

1. Build nghttp2 on windows with cmake which defines ssize_t as int. Do not build nghttp2 with NGHTTP2_NO_SSIZE_T because it does not work as you mentioned.
2. Write an application with new nghttp2_ssize API and build it with NGHTTP2_NO_SSIZE_T.  No ssize_t definition needed.

https://github.com/nghttp2/nghttp2/pull/2066 describes the motivation and its details.
We cannot change the definition of ssize_t to anything because we use it for years, and changing it might break the existing applications.

georgthegreat:
We build nghttp2 as a source-based monorepo development, hence we are unable to materialize the artifact (we even use our own build system).
I am not sure of any other build system is capable of defining CFLAG only for clients (afaik, bazel does not have this capability).

Maybe we can simplify the things to make the solution more straight forward?
I do not mind if actual some other type will be used for ssize_t on Windows (int is fine for me).

tatsuhiro-t:
If you use your own build system, I think you can define ssize_t to whatever type you want for systems that lack ssize_t, like windows.

I can add new macro to remove the code that uses ssize_t altogether so that nghttp2 can compile without it if it helps.

georgthegreat:
We can indeed, keep our own fork with arbitrary set of patches, we just do not like to maintain this fork, though this patch does not look very hard to maintain.

I do not quite understand of current behavior can be simplified without adding aditional flag.
We use static linkage hence we have no problems with ABI compatibility.

tatsuhiro-t:
You do not need a patch for this, define like "-Dssize_t=int" in cflags globally and compile everything.

georgthegreat:
global cflag would affect every command in every nghttp2 dependency disregarding of the module being compile is in any way related to it.

This would cause unpredictable breaks.

tatsuhiro-t:
Then we have no simple way to fix this.  This PR certainly breaks existing builds that expect ssize_t is int for windows.
I think you can build nghttp2 in shared library with its own cflags, and build the other parts of the program with their own cflags and linking it should work.

PR #2297, opened 2025-01-20
fix: Add defensive bounds checking in hd_ringbuf_init()

thevilledev:
Add bounds checking in hd_ringbuf_init() to prevent potential integer overflow during size calculations. While HPACK decoder controls its own buffer size (4-8K typical) and is not vulnerable to remote exploitation, this adds defensive programming guards for robustness.

### DIFF
--- a/lib/nghttp2_hd.c
+++ b/lib/nghttp2_hd.c
@@ -600,7 +600,7 @@ static int hd_ringbuf_init(nghttp2_hd_ringbuf *ringbuf, size_t bufsize,
     return NGHTTP2_ERR_NOMEM;
   }
 
-  for (size = 1; size < bufsize && size < max_size / 2; size <<= 1)
+  for (size = 1; size < bufsize; size <<= 1)
     ;
 
   if (size * sizeof(nghttp2_hd_entry *) > max_size) {
--- a/lib/nghttp2_hd.c
+++ b/lib/nghttp2_hd.c
@@ -594,9 +594,8 @@ static void hd_map_remove(nghttp2_hd_map *map, nghttp2_hd_entry *ent) {
 static int hd_ringbuf_init(nghttp2_hd_ringbuf *ringbuf, size_t bufsize,
                            nghttp2_mem *mem) {
   size_t size;
-  size_t max_size;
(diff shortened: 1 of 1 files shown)


lib/nghttp2_hd.c
```diff
@@ -594,8 +594,20 @@ static void hd_map_remove(nghttp2_hd_map *map, nghttp2_hd_entry *ent) {
 static int hd_ringbuf_init(nghttp2_hd_ringbuf *ringbuf, size_t bufsize,
                            nghttp2_mem *mem) {
   size_t size;
-  for (size = 1; size < bufsize; size <<= 1)
+  size_t max_size;
+
+  max_size = SIZE_MAX / sizeof(nghttp2_hd_entry *);
```
tatsuhiro-t, line 599:
```suggestion
  const size_t max_size = SIZE_MAX / sizeof(nghttp2_hd_entry *);

```
Add const because this is compile time constant.

lib/nghttp2_hd.c
```diff
@@ -594,8 +594,20 @@ static void hd_map_remove(nghttp2_hd_map *map, nghttp2_hd_entry *ent) {
...
+  if (bufsize > max_size) {
+    return NGHTTP2_ERR_NOMEM;
+  }
+
+  for (size = 1; size < bufsize && size < max_size / 2; size <<= 1)
```
tatsuhiro-t, line 604:
```suggestion
  for (size = 1; size < bufsize; size <<= 1)
```

Due to the definition of max_size and L600, size never overflow in this for loop.  No need to check `size < max_size / 2` in every iteration.

lib/nghttp2_hd.c
```diff
@@ -594,8 +594,20 @@ static void hd_map_remove(nghttp2_hd_map *map, nghttp2_hd_entry *ent) {
...
     ;
+
+  if (size * sizeof(nghttp2_hd_entry *) > max_size) {
```
tatsuhiro-t, line 607:
```suggestion
  if (size > max_size) {
```

thevilledev: true, `max_size` already accounts for `sizeof(nghttp2_hd_entry *)`

tatsuhiro-t:

PR #2303, opened 2025-02-02
Add QNX Support

dchordia:
Hi Tatsuhiro,
This contribution adds cross-compilation support for QNX operating system to nghttp2.
 
### Motivation
QNX is an industry standard real-time operating system for embedded systems, particularly for vehicles, featuring safety focused features. It has also recently been made free for non-commercial use through the QNX Everywhere program, aiming for students, researchers and hobbyists to experiment with the OS.
 
Given that nghttp2 is a mature HTTP/2 implementation with easy to use APIs as well as supporting applications, it would be great to enable nghttp2 to run natively on free version of QNX OS.

### Changes
- Adds a sys/time.h header to be included in util.h if building for QNX
- Adds conditionals on test build commands, omitting add_test() when building for QNX. Reason being we cross compile the library from a Linux Host and then transfer essential files to QNX target system, since we do not support self-hosted compilation toolchain for QNX yet.

### How to build
The build files and instructions for cross compiling nghttp2 for QNX are present at [qnx-ports](https://github.com/qnx-ports/build-files/tree/main/ports/nghttp2). A free QNX license can be obtained [here](https://www.qnx.com/products/everywhere/)

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -57,6 +57,9 @@
 #include "template.h"
 #include "network.h"
 #include "allocator.h"
+#if defined(__QNX__)
+  #include <sys/time.h>
+#endif
 
 namespace nghttp2 {
 
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -25,14 +25,22 @@ set(MAIN_SOURCES
   munit/munit.c
 )
 
+if(QNX)
+add_executable(main
+  ${MAIN_SOURCES}
+)
+else()
 add_executable(main EXCLUDE_FROM_ALL
   ${MAIN_SOURCES}
 )
+endif()
 target_link_libraries(main
   nghttp2_static
 )
-add_test(main main)
-add_dependencies(check main)
+if(NOT QNX)
+  add_test(main main)
+  add_dependencies(check main)
+endif()
 
 if(ENABLE_FAILMALLOC)
   set(FAILMALLOC_SOURCES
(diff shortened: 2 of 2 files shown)


tatsuhiro-t:
Instead of modifying tests/CMakeList.txt just for QNX, specify -DBUILD_TESTING=OFF to disable testing for a cross build.

dchordia:
I have added it to allow developers to test the library on their system when building for QNX as well.  Users can easily test by transferring libraries and binaries to target and running them. I am open to an alternative, but this seems ideal way to allow for testing on QNX.

src/util.h
```diff
@@ -57,6 +57,9 @@
 #include "template.h"
 #include "network.h"
 #include "allocator.h"
+#if defined(__QNX__)
```
tatsuhiro-t, line 60:
```suggestion
#ifdef __QNX__
```

src/util.h
```diff
@@ -57,6 +57,9 @@
...
+  #include <sys/time.h>
+#endif
```
tatsuhiro-t, line 62:
```suggestion
#endif // __QNX__
```

tatsuhiro-t:
Sadly, cmake is very limited in this area and no good way to just build tests.  The current way is a bit too complicated by adding bunch of QNX variable.
If you drop changes under tests directly, then I can merge this PR.

dchordia:
Okay. I have created a new PR with the appropriate requested changes. Closing this one in the others' favour.

PR #2305, opened 2025-02-05
Add QNX Support

dchordia:
Adding QNX cross compilation support. Based on feedback from PR #2303


tatsuhiro-t:

PR #2342, opened 2025-05-01
Fix CMake existence check for GetTickCount64

zjturner:
The CMake check for `GetTickCount64` on Windows only includes `<sysinfoapi.h>`.  If you check MSDN, it does say that `<sysinfoapi.h>` is the correct header, but it *also* says (include Windows.h).  Presently, the current check will fail 100% of the time because if you include sysinfoapi.h without including windows.h it's a compilation failure.  But if you do include windows.h, then it is not necessary to include sysinfoapi.h.

So the fix here is to simply change sysinfoapi.h to windows.h.  I confirmed after this patch that `HAVE_GETTICKCOUNT64` will be 1 on Windows, whereas before it would be undefined.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -369,7 +369,7 @@ check_function_exists(clock_gettime HAVE_CLOCK_GETTIME)
 check_function_exists(mkostemp  HAVE_MKOSTEMP)
 check_function_exists(pipe2     HAVE_PIPE2)
 
-check_symbol_exists(GetTickCount64 sysinfoapi.h HAVE_GETTICKCOUNT64)
+check_symbol_exists(GetTickCount64 "windows.h;sysinfoapi.h" HAVE_GETTICKCOUNT64)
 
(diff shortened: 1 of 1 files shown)


zjturner:
yes.  I used `--debug-trycompile` so that it would leave the generated try compile build system on disk.  I cd'ed into the directory and typed `ninja` and it was failing to compile due to something in `winnt.h` saying that no architecture was defined.

You can actually reproduce this problem yourself, independently of cmake or any external build system.  Make an empty file and just add `#include <sysinfoapi.h>` to it.  No other code.  Run `cl.exe` against it with `/D_WIN32` and whatever other options you want, and you'll get the same error.   Change to `#include <windows.h>` and it will work.

tatsuhiro-t:
Thank you.  So we included windows.h in configure.ac and it worked fine, but cmake didn't.

tatsuhiro-t:

PR #2346, opened 2025-05-05
Exclude [, ], and # from allowed :path characters

islandryu:
Refs: https://github.com/nodejs/node/issues/57988

This change excludes `[`, `]`, and `#` from valid `:path` characters.  
According to [rfc 91113](https://datatracker.ietf.org/doc/html/rfc9113#section-8.3.1-2.4.1), yhe `:path` field consists of `path-absolute`, `"?"`, and `query`, 
These characters are not part of `path-absolute` or `query` in [rfc 3986](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A). 
```
query         = *( pchar / "/" / "?" )
path-absolute = "/" [ segment-nz *( "/" segment ) ]
segment     = *pchar
segment-nz    = 1*pchar
pchar       = unreserved / pct-encoded / sub-delims / ":" / "@"
unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
sub-delims  = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="
```


tatsuhiro-t:
This check has been added by 83c063346d9ff7a4f449010c6f0460ef7171e5ef, and those characters are explicitly allowed for compatibility reasons.
I would like to keep the current behavior as is because tightening this kind of check most likely causes troubles and gets some push backs.

islandryu:
Thx, I appreciate the context.
I'll close the PR.

PR #2455, opened 2025-06-08
nghttpd: Support SSLKEYLOGFILE

starrify:
Maybe relevant: https://github.com/nghttp2/nghttp2/pull/2119

### DIFF
--- a/src/HttpServer.cc
+++ b/src/HttpServer.cc
@@ -2236,6 +2236,11 @@ int HttpServer::run() {
       return -1;
     }
 #endif // NGHTTP2_OPENSSL_IS_BORINGSSL && HAVE_LIBBROTLI
+
+    if (tls::setup_keylog_callback(ssl_ctx) != 0) {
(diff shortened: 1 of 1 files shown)


src/HttpServer.cc
```diff
@@ -2236,6 +2236,11 @@ int HttpServer::run() {
       return -1;
     }
 #endif // NGHTTP2_OPENSSL_IS_BORINGSSL && HAVE_LIBBROTLI
+
+    if (tls::setup_keylog_callback(ssl_ctx) != 0) {
+      std::cerr << "Failed to setup keylog" << std::endl;
+      exit(EXIT_FAILURE);
```
tatsuhiro-t: You should return -1 instead of calling exit to align the convention done in this function.

starrify: Thanks a lot! The branch has been updated accordingly.

tatsuhiro-t:

PR #2466, opened 2025-06-27
Provide fallback ssize_t type choice as SSIZE_T on MSVC

cosmo0920:
MSVC has already a special type of SSIZE_T.
So, we need to use this.

This is because when using nghttp2 as static library with MSVC, we got:

```
nghttp2.h(981): error C2628: 'ptrdiff_t' followed by 'int' is illegal (did you forget a ';'?) 
```

After I use SSIZE_T in fallback mechanism instead,
cl.exe was able to compile in that part of nghttp2.

A system of Windows data types has `SSIZE_T`:

https://learn.microsoft.com/en-us/windows/win32/winprog/windows-data-types

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -340,7 +340,11 @@ check_type_size("ssize_t" SIZEOF_SSIZE_T)
 if(SIZEOF_SSIZE_T STREQUAL "")
   # ssize_t is a signed type in POSIX storing at least -1.
   # Set it to "int" to match the behavior of AC_TYPE_SSIZE_T (autotools).
-  set(ssize_t int)
+  if (MSVC)
+    set(ssize_t SSIZE_T)
+  else()
+    set(ssize_t int)
+  endif()
 endif()
 # AC_TYPE_UINT8_T
 # AC_TYPE_UINT16_T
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
That breaks the existing projects that use int as ssize_t.
I am not sure how you get error like: ptrdiff_t' followed by 'int' is illegal.
Is nghttp2_ssize defined as int somewhere?

cosmo0920:
We use the following settings to build nghttp2 in Fluent Bit:

```
FLB_OPTION(ENABLE_LIB_ONLY   ON) #=> -DENABLE_LIB_ONLY=On
FLB_OPTION(ENABLE_STATIC_LIB ON) #=> -DENABLE_STATIC_LIB=On
FLB_OPTION(ENABLE_SHARED_LIB OFF) #=> -DENABLE_SHARED_LIB=Off
FLB_OPTION(BUILD_SHARED_LIBS OFF) #=> -DBUILD_SHARED_LIBS=Off
FLB_OPTION(BUILD_STATIC_LIBS ON) #=> -DBUILD_STATIC_LIBS=On
FLB_DEFINITION(NGHTTP2_STATICLIB) #=> -DNGHTTP2_STATICLIB and define #if def ... #endif block inside the C sources
```

> That breaks the existing projects that use int as ssize_t.

We're still investigating for how to solve such type collisions.
Providing a fallback option for SSIZE_T instead of int plus default fallback type is int is acceptable in this project?
This is because it's unnatural to fallback as int from ssize_t.

cosmo0920:
> Is nghttp2_ssize defined as int somewhere?

nghttp2_ssize is resolved as ptrdiff_t in our source tree.

cosmo0920:
After digging with git bisect, this issue is caused by our fault.
So, we was able to remote our specific patch here: https://github.com/fluent/fluent-bit/commit/4be9df40fdb41621145f548e5e898b8e7dd16eb5
It's not correct fix and just shadowing onigmo's config.h things.

PR #2470, opened 2025-07-03
lib: enable overriding NGHTTP2_INBOUND_BUFFER_LENGTH

rsaxvc:
Making this adjustable eases integration with microcontrollers and other small systems whose stacks may be much smaller than the default inbound buffer length. 

Supports espressif/idf-extra-components#523


Bryght-Labs-Richard:
Please let me know if there's anything I can do to help with this PR, or anything to do differently. It's really great to be able to use HTTP2 on RTOS systems.

PR #2473, opened 2025-07-23
Use allocator-aware free in failure path

geoffhill:
This change avoids [free(3)](https://man.openbsd.org/free.3) from stdlib in favor of `nghttp2_mem_free()` for freeing a buffer in an error path. The buffer is allocated with `nghttp2_mem_malloc()`.


tatsuhiro-t:

PR #2480, opened 2025-08-17
Make error handling robust

tatsuhiro-t:
Stream errors are now promoted to connection errors.  This means that an event that previously just resets a single stream now closes a connection entirely.  The promoted errors are mostly implementation errors.  Some involve HTTP fields, but they are already treated stream error.  People who care about that should have already raised any issues.  We do not have any outstanding related issues now, so it seems OK to treat it as connection error.

We have some contradictory specifications around
nghttp2_on_invalid_header and nghttp2_on_invalid_header2 callbacks. nghttp2_on_invalid_header says that if it is omitted, a stream is reset.  Meanwhile, nghttp2_on_invalid_header2 says that if it is omitted, invalid field is silently ignored.  In actual implementation, if both omitted, we treat it as stream error.  In practice, it is often required not to bail out if invalid header is received.  In this change, if both callbacks are omitted, invalid field is silently ignored as the documentation of nghttp2_on_invalid_header2 says.  The connection error promotion is applied here as well.  So if invalid field is received, and callback returns
NGHTTP2_ERR_TEMPORAL_CALLBACK_FAILURE, it is treated as connection error.

### DIFF
--- a/lib/includes/nghttp2/nghttp2.h
+++ b/lib/includes/nghttp2/nghttp2.h
@@ -2045,12 +2045,13 @@ typedef int (*nghttp2_on_header_callback2)(nghttp2_session *session,
  * The parameter and behaviour are similar to
  * :type:`nghttp2_on_header_callback`.  The difference is that this
  * callback is only invoked when a invalid header name/value pair is
- * received which is treated as stream error if this callback is not
- * set.  Only invalid regular header field are passed to this
- * callback.  In other words, invalid pseudo header field is not
- * passed to this callback.  Also header fields which includes upper
- * cased latter are also treated as error without passing them to this
- * callback.
+ * received which is treated as stream error if this callback returns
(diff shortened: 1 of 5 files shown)


aduh95:
This causes test failures and timeouts on Node.js, it would probably make sense to treat it as semver-major

thinred:
I would be very good to resolve it:

- The release of the new nghttp2 in Debian is now blocked on this, see https://qa.debian.org/excuses.php?package=nghttp2
- To go around this, the nodejs project has patched their nghttp2, see https://github.com/nodejs/node/pull/59790

Overall, I would prefer to avoid patching nghttp2 the way that it was done for nodejs. @tatsuhiro-t - any chance this can be addressed? what about the proposal in https://github.com/nghttp2/nghttp2/pull/2480#issuecomment-3368086811?

PR #2500, opened 2025-09-01
Added nghttp3's pattern targets

bmarques1995:
Now the nghttp2 install is following the same pattern of nghttp3, that can be seen on https://github.com/ngtcp2/nghttp3/blob/main/lib/CMakeLists.txt, introduced on [ngtcp2/nghttp3#228](https://github.com/ngtcp2/nghttp3/pull/228)


tatsuhiro-t:
@vszakats What do you think about this change?  Looks similar to nghttp3 PR, a little bit different in detail.

vszakats:
@tatsuhiro-t Agree, this is reasonable to do for nghttp2, and the patch looks good to me. (I've tested the install, but haven't the consuming side)

tatsuhiro-t:
Thank you for your review, @vszakats. Always appreciated.

bmarques1995:
> Thank you for your review, @vszakats. Always appreciated.

Guys, do you need any change?

tatsuhiro-t:

Issue #2501, opened 2025-09-02
nghttp2 overwrites Find* of the parent project, unusable with CPM


Anton3:
Current `CMakeLists.txt` says:

```cmake
set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake" ${CMAKE_MODULE_PATH})
```

The way nghttp2 prepends its own `Find*` modules instead of appending them is invalid on both practical and theoretical bases.

When I build nghttp2 as ExternalProject (e.g. using [CPM](https://github.com/cpm-cmake/CPM.cmake)), I expect being able to overwrite the way nghttp2 discovers packages by providing my own `Find*.cmake`.

For example, nghttp2 does `find_library` in its `FindLibev.cmake`, thus assuming that libev is installed. Meanwhile, one may want to provide an option of acquiring libev using CPM. Note that nghttp2 not knowing about CPM would not be an issue if nghttp2 allowed the parent project to override its `Find*` modules.

Generally speaking, when a parent project builds nghttp2 and another project (let's call it X) in-tree, and their `Find*` conflict, it should be the parent project's right and responsibility to override `Find*` to ensure that the needs of both projects are satisfied.

Another point is that if nghttp2 is built as ExternalProject, the parent project may also want to use Libev, and may have already called `find_package(Libev)` using its own `Find*`. Then if nghttp2 insists on using its own `Find*`, then we may have a linker error, as one Libev may come from e.g. Conan, and another one from a system package.

See some previous discussion on PREPEND vs APPEND: https://github.com/ros/cmake_modules/issues/21

PR #2502, opened 2025-09-02
fix cmake: append to CMAKE_MODULE_PATH instead of prepending

Anton3:
Fixes: #2501

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,12 @@ set(LT_CURRENT  42)
 set(LT_REVISION 5)
 set(LT_AGE      28)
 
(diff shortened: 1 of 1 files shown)


Anton3:
At a glance, Gemini states a valid concern.

Thinking about it further, `APPEND` should be fine as long as all child `ExternalProject`s use `APPEND`.

The root project starts with an empty `CMAKE_MODULE_PATH` (system `Find*` use other mechanisms). Anything anywhere in `CMAKE_MODULE_PATH` will override those system modules.

If a parent project or a "sibling" project has already modified `CMAKE_MODULE_PATH` to use its own `Find*` modules, then they presumably work, and continuing to prefer our own `Find*` modules would only lead to linker errors if we find different packages from them.

PR #2558, opened 2025-11-10
src: Rewrite DList

tatsuhiro-t:
Rewrite DList, and now is named SList because it is Singly Linked List.


tatsuhiro-t:
While it is fun to explore C++ features, SList is less efficient than DList.  Close without merging this.

PR #2579, opened 2025-12-07
nghttpx: Remove stream_closed_ from Http2DownstreamConnection

tatsuhiro-t:
Now RST_STREAM handling is improved in libnghttp2, we can just submit RST_STREAM, and let lbnghttp2 decide whether the frame should be sent or not.


tatsuhiro-t:
@gemini-code-assist nghttp2_submit_rst_stream does not return error if the specified stream is not found.

PR #2677, opened 2026-04-12
OpenSSL 4.0.0 beta1 fix

xl32:
Dear @tatsuhiro-t,

This is the fix to build ngtcp2 against OpenSSL 4.0.0 beta 1.
Please review.

### DIFF
--- a/src/shrpx_tls.cc
+++ b/src/shrpx_tls.cc
@@ -2357,7 +2357,7 @@ ssize_t get_x509_fingerprint(uint8_t *dst, size_t dstlen, const X509 *x,
 }
 
 namespace {
-std::string_view get_x509_name(BlockAllocator &balloc, const X509_NAME *nm) {
+std::string_view get_x509_name(BlockAllocator &balloc, auto *nm) {
   auto b = BIO_new(BIO_s_mem());
   if (!b) {
     return ""sv;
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Thank you for PR.  It seems like this does not work with wolfssl.  Perhaps, boringssl also fails.  We should also check that these ASN_STRING_length and ASN_STRING_get0_data available for libressl.

src/shrpx_tls.cc
```diff
@@ -2357,7 +2357,7 @@ ssize_t get_x509_fingerprint(uint8_t *dst, size_t dstlen, const X509 *x,
 }
 
 namespace {
-std::string_view get_x509_name(BlockAllocator &balloc, X509_NAME *nm) {
+std::string_view get_x509_name(BlockAllocator &balloc, const X509_NAME *nm) {
```
tatsuhiro-t:
```suggestion
std::string_view get_x509_name(BlockAllocator &balloc, auto nm) {
```

This fixes the compile error with wolfssl.  Could you check that this also works with openssl 4 beta?

xl32: Yes, it works fine and passes tests. Thanks, PR updated.

tatsuhiro-t:

PR #2681, opened 2026-04-15
Make header related errors stream error with glitch rate limit guard

tatsuhiro-t:
See #2604


pimterry:
There's one remaining HTTP message case that's not covered here I think: data size > expected content length.

This reverts the connection error to a stream error for the inverse case (where you receive less data than expected) but keeps it as a connection error for too-much-data. Here: https://github.com/nghttp2/nghttp2/blob/0e11ace645adc09f4f1581986c345a924185c6db/lib/nghttp2_session.c#L6783-L6785

Is that intentional? I think it'd be better to make the too-much/too-little cases consistent, and imo both are HTTP message-level errors that should fail at the stream level.

tatsuhiro-t:
Fixed by 3a2ad6e

PR #2773, opened 2026-05-25
src: Avoid zero-argument std::println

kmehltretter82:
The zero-argument overload of std::println is C++26.  Use an empty format string so these line-separator prints remain valid with C++23 standard library implementations.

This fixes a build failure with Apple clang 16.0.0 (clang-1600.0.26.6) and the macOS 14.5 SDK, whose libc++ provides std::println(format, args...) but not std::println().


tatsuhiro-t:

PR #2811, opened 2026-07-07
Restore stream_id==0 validation for PRIORITY frames

fidgetaryan445:
RFC 9113 section 5.3.2 deprecates the `priority signaling`, but in the same RFC 9113 in section 6.3 writes `The PRIORITY frame always identifies a stream. If a PRIORITY frame is received with a stream identifier of 0x00, the recipient MUST respond with a connection error (Section 5.4.1) of type PROTOCOL_ERROR.

If not resolved, this could lead to a delayed detection of a connection that should have been terminated on arrival of a buggy `PRIORITY`  frame. I believe that commit `3dd61f8` was an aggressive one and might have ignored this.


hiSandog:
Restoring the stream-zero check matches the explicit PRIORITY-frame requirement in RFC 9113. In addition to the direct receive test, a useful edge case would feed a stream-zero PRIORITY frame after another valid frame in the same input buffer and verify that processing stops with a connection-level `PROTOCOL_ERROR`. That would confirm the invalid frame cannot be treated as merely ignored priority signaling while subsequent buffered frames continue.

PR #2828, opened 2026-08-09
util: add <fstream> include

thesamesam:
This shows up with GCC 17 where some transitive internal includes changed: it worked by chance before.


trofi (approved):

tatsuhiro-t:

